PR #22, opened 2020-08-05
Added more torrent statuses and removed the ability to select the text in the torrents themselves

tmayoff:
Since most of my already existing torrents were showing up as failed I fixed that. Also I fixed the fact that you could select the text inside the torrents, which was annoying me


WDaan:
Thank you very much for contributing 🙂

PR #45, opened 2020-10-15
Improvements on AddModal

muertocaloh:
- Added a checkbox in the add torrent modal to allow users that upload content skip the file hash check

- Set as false the autoTMM param to download the torrent into the specified path

- Added a combobox to select the torrent category on adding

image


WDaan:
Thanks you very much for you contribution! Looks great! 
It almost looks like I've written it myself

PR #87, opened 2020-12-30
mobile devices spacing issues

colthreepv:
Background story: this project is really useful to control qbittorrent from a smartphone

Mine has 393x713 px viewport, I bet most have similar but not perfectly equal viewport size  
My changes aim to fix display for most of those resolutions, mostly 380px in width and beyond

# Fixes
- Less margins and paddings on Extra Small devices for Dashboard component
- Custom CSS to fix navbar buttons appearance

I know that there was **zero** custom CSS before this Pull Request, but I also think mobile device usability is really important.  
If you don't like how I made the files or anything, it can be changed

Resolution: 380 x 700
**BEFORE** (origin/master as now)
localhost_8000_

**AFTER PR**
localhost_8000_ (1)


WDaan:
Thanks for taking an interest!
Your fix works for a few more resolutions, but indeed still breaks on smaller screens.
image

Also, if I were to add another option up there, it's broken again.

I think I should implement some kind of fab/fly-out menu for the smaller screens.

colthreepv:
I agree, but still I think the lowest resolution to aim for should be 360px width, not less (that should be a Galaxy S5)

WDaan:
Atm it's allright to use up to Galaxy S5... When the need arises for extra functionality, I'll figure out a better solution

PR #129, opened 2021-01-29
Fixed white line below pagination that is visible in dark theme mode

IevgenSobko:
# Fixes
- Fixed white line below pagination that is visible in dark theme mode

Before

After

Also verified for mobile view. Looks OK.


WDaan:
Thanks!

PR #131, opened 2021-01-29
Fixed forceDL misspeling that caused forced downloading torrents to be displayed as Failed

IevgenSobko:
# Fixes
- fixed forceDL misspelling that caused forced downloading torrents to display as Failed
# Improvements
- added identification for forced for both downloading and uploading torrents

### DIFF
--- a/src/models/Torrent.js
+++ b/src/models/Torrent.js
@@ -35,12 +35,14 @@ export default class Torrent {
 
   formatState(state) {
     switch (state) {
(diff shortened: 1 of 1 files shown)


WDaan:
Doesnt the identification break the color? As it's chosen based on state name

IevgenSobko:
I haven't noticed color breakage
I have changed just formated string returned from function not actual names I guess, but I might be wrong.

PR #145, opened 2021-02-06
feat: add uploaded property

jef:
# Features

Adds `Uploaded` property to torrent and changes ordering.

uploaded

# Improvements

Adds `QBITTORRENT_PORT` and `VUETORRENT_PORT` for development. Defaults to original ports, but if developer has different ports, this is useful without changing code.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -1,4 +1,8 @@
 const webpack = require('webpack')
+
+const qBittorrentPort = process.env['QBITTORRENT_PORT'] ?? 8080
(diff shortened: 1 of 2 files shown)


WDaan:
Great catch! It must've gotten lost in a merge somewhere

vue.config.js
```diff
@@ -27,10 +31,10 @@ module.exports = {
       poll: true
     },
     host: '0.0.0.0',
-    port: 8000,
```
jef, line 30: Not sure why I didn't just set this to the value...

```suggestion
    port: vueTorrentPort,
```

Whoops :sweat_smile:

PR #169, opened 2021-02-27
feat: current speed cards redesign

IevgenSobko:
Current speed cards redesign

# Improvements
- current speed cards redesign

### DIFF
--- a/src/components/Core/SpeedCard.vue
+++ b/src/components/Core/SpeedCard.vue
@@ -1,29 +1,31 @@
 <template>
   <v-card
     flat
+    rounded="md"
     color="secondary"
     class="speedCard"
     data-testid="SpeedCard"
   >
-    <v-layout row :class="color + '--text'">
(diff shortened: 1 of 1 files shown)


IevgenSobko:
Following issue https://github.com/WDaan/VueTorrent/issues/141 I created a pull request to not lost this design.
In my opinion, it is better than current and follows current conventions with rounded corners for cards in the side panel.
I have seen someone liked my pill design but it will be a little bit off the current design. Anyway, after this change, we can change to pill design by changing just one property to `pill` value

WDaan:
First of all, thanks a lot for helping me out :) 

I'm not a designer by any means but I think all properties (icon, numbers & units) should be either all horizontal or all vertical.
I personally like the first design you've made in that issue....even though there is a lot of 'wasted' space on the sides.

IevgenSobko:
All being vertical takes a lot of space as you mentioned in the issue.
So I redesigned it to take less... :)

PR #171, opened 2021-02-28
feat: improved per-torrent speed limits modal

IevgenSobko:
# Fixes
- long strange floating number on initialization when speed was unlimited
# Improvements
- autoselecting value on modal open so you can immediately start setting correct limit without moving the mouse to it 
- firing submit on `Enter` in the text field


IevgenSobko:
Oh seems PR picked up way too many files than I expected.
I'll try to open it again without conflicts

PR #173, opened 2021-02-28
feat: torrent detail infinity sign for unlimited speeds

IevgenSobko:
# Features
- added infinity sign instead of None for unlimited speeds in torrent detail modal
# Fixes
- space in units for upload/download speed limit in torrent info modal


IevgenSobko:
@WDaan you also might want to refactor your changes in `src/filters.js` that were necessary for speed limits implementation that returned `None`.
In particular `getDataValue` and `getDataUnit` functions...

PR #187, opened 2021-04-02
AddTrackers dialog bug fix

m4ximuel:
Does not close after clicking close. There is a problem with the declaration.

# Fixes
- AddTrackers dialog bug fix

### DIFF
--- a/src/components/Modals/TorrentDetailModal/Tabs/Trackers.vue
+++ b/src/components/Modals/TorrentDetailModal/Tabs/Trackers.vue
@@ -48,7 +48,7 @@
       >
         Delete
       </v-btn>
-      <v-dialog v-model="dialog" persistent max-width="290">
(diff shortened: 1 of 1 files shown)


WDaan:
I can't seem to reproduce this?  For me the cancel button works just fine?

image

m4ximuel:
> I can't seem to reproduce this? For me the cancel button works just fine?
> 
> image

Existing source code shares the same function 'addTrackers' for 'cancel' and 'add'. And the definition for 'dialog' is nested with the parent.

WDaan:
Yes, you're absolutely right! The close behavior isn't as it should be.

The nested 'dialog' shouldn't be a problem? It's generated from a mixin, aren't these unique?

m4ximuel:
Vuejs DevTools gave me a nested warning about 'dialog'.
I don't think it will be a problem even if we don't fix it, but it's uncomfortable.

PR #189, opened 2021-04-03
Update mutations.js

m4ximuel:
I haven't yet understood the role for 'rid'.
However, I know that increasing the number of 'rid' hide the 'available tag lists'.

# Fixes
- After browsing the'Torrent Details' of several torrents, the list of available tags is hidden.

### DIFF
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -53,8 +53,9 @@ export default {
     state.authenticated = payload
   },
   updateMainData: async state => {
-    const rid = state.rid ? state.rid : undefined
-    const res = await qbit.getMainData(rid)
+    //const rid = state.rid ? state.rid : undefined
+    //const res = await qbit.getMainData(rid)
+    const res = await qbit.getMainData(undefined)
 
     // status
     state.status = new Status(res.data.server_state, res.data.tags)
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -53,8 +53,6 @@ export default {
     state.authenticated = payload
   },
   updateMainData: async state => {
-    //const rid = state.rid ? state.rid : undefined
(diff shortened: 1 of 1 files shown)


WDaan:
it's in the [docs](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)) :smile: 

image

m4ximuel:
How about keeping full_update by setting 'undefined' or 'true' to updateMainData? There seems to be a bug about the data being only partially updated.
All data displayed in the drawer and the 'list of available tags' often loses all data until they are refreshed.

WDaan:
Fine by me :slightly_smiling_face:

m4ximuel:
ok, Keep the comments for now. 
Let's write the logic to update the data by only partially updating the data later.

WDaan:
I'm not sure I follow?
Why not reload all data? It's more reliable & probably barely introduces additional delay

m4ximuel:
There are countless people in this world, some of whom may manage thousands of torrents.
😅

WDaan:
Yeah I always tend to forget about those 😅🤣 my bad

m4ximuel:
Thanks, I'll check the api docs and study it.
If you have time, please check this #186
I am confident that it will make our lives richer.

WDaan:
Many thanks for taking an interest in this project! :smile: 
It's nice to have an extra set of eyes and opinion on things instead of me just winging it.

I've looked at it & it looks really promising! I think a lot of people will benefit from that.
I do have some harware issues atm, so I haven't been able to test it out myself

m4ximuel:
I think there will be no problem because the code has been tested offline several times.
I hope you release it later by modifying the '//toast' part as you like.

PR #194, opened 2021-04-04
Update Dashboard.vue

m4ximuel:
Layout improvements
changed to show as draggable items


WDaan:
Awesome! This is a great improvement, thanks!

PR #195, opened 2021-04-04
Update TorrentDetailModal.vue

m4ximuel:
Please consider this fun feature. It will often be useful to desktop users who use subtly small sizes.

# Features
- Add fullscreen button for desktop users


WDaan:
Great! I like it :slightly_smiling_face:

PR #196, opened 2021-04-05
Feedback to the user based on the network response.

m4ximuel:
# Features
- When a network problem occurs during operation, show dialog included text 'waiting for response'
- If a 403 error code is returned, logout is processed.
# Fixes
- torrent list is no longer be freeze on device that has been left unattended in sleep mode for a long time or device with a network error.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -10,6 +10,26 @@
     <v-main class="background">
       <router-view />
     </v-main>
+    
+    <v-dialog
+      v-model="networkErrDialog"
+      persistent
+      width="400"
+      content-class="noselect"
+    >
+      <v-card color="primary">
+        <v-card-text>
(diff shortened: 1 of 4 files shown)


WDaan:
Hi!

Why all the sudden closes? I really do like the features you bring, but some are harder to test out than others... and I haven't gotten around to it yet :slightly_smiling_face: 

Also please make a PR for every new feature or every fix seperatly.... that way they are easier to test/fix & merge :slightly_smiling_face:

m4ximuel:
I have the original source locally. It's my first time working on GitHub, so I'm working on it for a while while studying. I'll commit again with your new changes.

WDaan:
No worries at all! You're doing great & it's really good to practice your git skills

PR #201, opened 2021-04-07
Update ConfirmDeleteModal.vue

m4ximuel:
# Fixes
- not release selected list when close dialog without close button


WDaan:
Yes, great catch!

PR #203, opened 2021-04-09
Validate the new mainData

m4ximuel:
If the new value is not directly compared to undefined, the new value is not applied to the vuetorrent when the server returns a value of false.
Example) When altspeed is false

### DIFF
--- a/src/models/Status.js
+++ b/src/models/Status.js
@@ -12,18 +12,24 @@ export default class Status {
   }) {
 
     const previous = store.state.status
-
-    this.status = connection_status || previous.status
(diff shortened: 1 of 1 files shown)


WDaan:
So it only happens for altspeed? Why the changes for every property instead of a simple:
```
this.altSpeeds = use_alt_speed_limits === undefined ?  use_alt_speed_limits : previous.altSpeed
 ```

m4ximuel:
The requirements for operation are all the same, so I modified it like that for now.
From a performance optimization point of view, modifying altspeed alone is not bad either.

Issue #59, opened 2020-10-26
Consistent height for Torrent Detail modal


EduCampi:
The Torrent Detail modal currently switches height depending on which tab is being looked at.

It would be good to have some consistency, that way one can navigate to the sides whiteout having to chase up and down with the mouse.

On top of that, the 'Info' and 'Content' tabs seem to be too small for their content. Right now it occupies around 25% of the height of my browser window, while the remaining 75% is wasted space.
That also makes it take too much time to navigate the content tab for torrents with many files, as we have to scroll a lot.

Would it be possible to fix the max/min height to the same value?

If you don't agree with having the size constant for all tabs, maybe at least change the Content one to make file navigation easier.

Maybe changing
components/Modals/TorrentDetailModal/Tabs/Content.vue
v-card-text style
so instead of '500px' is something like 60, 70 or 80vh ?

Maybe being in px is not scaling properly to my screen resolution, it seems like taking advantage of something like 70% of the viewport size would be best, as we are not interacting with the rest of the page when looking at the modal.

Thanks,

PR #207, opened 2021-04-10
Fix layout issues and more

m4ximuel:
# Fixes
- Fix layout issues
> - Fixes an issue where 'torrent details' modal does not work properly on iOS
> - 'Vue2-perfect-scrollbar' is no longer dependent
> - A few pixels at the end of the modal's rounded corners no longer show the background color
> - When more than one modal is open, it prevents keyboard input from the dashboard

# Improvements
- You can adjust the roundness of the entire modal through the 'form' and 'formtop' elements in 'src/styles/variables.scss'.
> - 'formtop' must be defined as a larger size than 'form'. This is how to solve the problem of displaying the background color by overlapping the rendering alpha value.
- The entire 'dashboard' and the entire 'add modal' area now respond to file drag and drop.
- now can add files by dropping them in the entire area with splash screen.


src/components/Modals/ChangeLocationModal.vue
```diff
@@ -81,6 +76,19 @@ export default {
     },
     torrent() {
       return this.getTorrent(this.hash)
+    },
+    isPhone() {
+      return this.$vuetify.breakpoint.xsOnly
+    },
+    dialog: {
```
WDaan, line 83: Why copy-paste this code for every single modal? That's the whole point of the Modal Mixin

src/components/Modals/SettingsModal/SettingsModal.vue
```diff
@@ -1,76 +1,75 @@
...
   <v-dialog
     v-model="dialog"
+    max-width="800px"
     scrollable
+    :content-class="isPhone ? 'rounded-0' : 'rounded-form'"
     :width="dialogWidth"
-    :fullscreen="phoneLayout"
+    :fullscreen="isPhone"
   >
-    <v-card style="min-height: 400px; overflow: hidden !important">
```
WDaan, line 8: I'm not sure about removing the dedicated min-height... on desktop the modal height now jumps around on different tabs and that's pretty annoying. Issues were opened in the past for this very issue.
Same goes for other modals as well

src/components/Modals/SearchModal/SearchModal.vue
```diff
@@ -99,6 +99,7 @@ export default {
   mixins: [Modal, FullScreenModal, General],
   data() {
     return {
+      hndlDialog: true,
```
WDaan, line 102: You forgot to remove the perfect-scrollbar in this modal

src/main.js
```diff
@@ -20,10 +20,6 @@ Vue.use(Toast, {
   timeout: 2000
 })
 
-import PerfectScrollbar from 'vue2-perfect-scrollbar'
```
WDaan, line 23: You forgot to remove it in both SeachModal.vue & TorrentDetailModal > Peers.vue
It can also be removed as a dependency from package.json :)

src/components/Torrent/Torrent.vue
```diff
@@ -1,15 +1,15 @@
...
   <v-card
-    flat
-    class="pointer noselect"
-    :class="{ selected: isSelected}"
-    @click.native.exact.prevent="selectMode && selectTorrent(torrent.hash)"
-    @dblclick.prevent="showInfo(torrent.hash)"
-    @click.ctrl.exact.prevent="selectTorrent(torrent.hash)"
-    @click.shift.exact.prevent="selectUntil(torrent.hash, index)"
+    class="pointer noselect elevation-0 rounded-0 ma-0 pa-0"
+    :class="isSelected ? 'info' : ''"
```
WDaan, line 4: I really dislike the flashy blue select color

src/mixins/Modal.js
```diff
@@ -2,15 +2,7 @@ import { mapGetters } from 'vuex'
 export default {
   props: ['guid'],
   computed: {
-    ...mapGetters(['getModalState']),
```
WDaan, line 5: I still think a Mixin is the best way to have common logic instead of copy-pasting the same code in every modal... thats not easily maintainable. I would prefer improving this mixin instead.

src/views/Dashboard.vue
```diff
@@ -18,170 +22,167 @@
...
+      <v-expand-x-transition>
+        <v-card
+          v-show="searchFilterEnabled"
+          id="searchFilter"
+          flat
+          xs7
+          md3
+          class="ma-0 pa-0 mt-1 transparent"
+        >
+          <v-text-field
```
WDaan, line 35: This new search is pretty cool!! :)

WDaan (changes requested):
Great work on the modal re-do! They look very slick.
I do have a few small remarks, please do not take these personal

Also please always restart from master when making changes instead of building on top of your current changes.
Such big PR's with lots of changes are insanely hard to review.

m4ximuel:
delete modal not working, I'll see all

src/components/Modals/ChangeLocationModal.vue
m4ximuel: because of closing animation.

WDaan: Yes, just increase the delete timeout a little bit in the mixin. This copy/pasting for the dialog introduces a lot of extra boiler plate code

m4ximuel:
@WDaan Please check the changes.
And please tell me the color you want for the selected torrent. Colors must be specified for future compatibility reasons.

WDaan:
I will! As soon as I can, probably tomorrow morning.

I don't know, what was wrong with the old color? Not distinctive enough?

m4ximuel:
The menu itself was behaving incorrectly. In addition, even though the one-click menu is not actually selected, it is expressed in the 'selected color', causing confusion to the user, so I removed the function that the menu is basically 'selected'. Therefore, the torrent list now need a 'selected color' for defined to maintain correct compatibility.

m4ximuel:
context-menu now perfect.
Press and hold with one finger to open
It opens instantly when pressed with two fingers.
----
Two-finger touch reacts only when touching the same item, 
It is designed in consideration of various situations and respects the user's intention as the top priority.

WDaan:
Awesome! Once I have a look at it, we can merge & release :)

m4ximuel:
@WDaan Icons and other details do I put off to the next brunch?
There are many changes, so I think it is good to release it this time.

WDaan:
Yes, please do make a seperate PR for those. This one is little too big allready

src/components/Modals/SettingsModal/SettingsModal.vue
m4ximuel: It doesn't seem to be reproduced after fixing the layout.

WDaan:
I'm looking into using maybe using [vue-js-modal](https://www.npmjs.com/package/vue-js-modal) for modals, like it's used in [this](https://github.com/ekvedaras/redis-gui) project here. This may or may not be a good way to do it, but I think it's worth trying out.

Otherwise the Modal mixin should still be refactored a little bit, so we don't have the dialog handler code for evey single dialog

m4ximuel:
@WDaan It's really cool, but we need to review it.
Our problem is that when we issue the modal remove command on the mutation, the modal gets removed before showing the close animation.
Therefore, I'm studying possible solutions within Modal.js, respecting your opinion of disagreeing with the repeat of the current source code. I also agree it's not good way.
Returning to your opinion, the Vue.js modal is really interesting. I think this will make the 'torrent details' more useful. As for the Vue.js modal, how about testing it in a new brunch after this brunch is finished?
The current brunch will be fetched after Modal.js is improved.

WDaan (commented):
Looks great! :)

What do you think about general modal height for both settings/info modal? Now the height jumps around when changing tabs, which some people find a annoying?-

WDaan:
FYI the new right click menu is AMAZING, works like charm on both desktop & mobile!
Really great job on that one! :tada:

m4ximuel:
@WDaan I think the modal height change is annoying because it hides sometimes the [SAVE] button in the past. However, the current version fixes that problem. and,  I think that the modal height changes proportionally to the content when the tab changes is the standard that vuetify.js pursues. If the problem is that the content height is too short and the modal size variation is large, I agree on min-height on this point. What were people dissatisfied with?

WDaan:
Issue: #59  
It's not that a lot of people complained (VueTorrent wasn't this popular back then either) but I do agree with some of the points made. 

Maybe the size can vary indeed.... but it shoudln't jump around THAT much. I think re-introducing a minimum height, isn't a bad idea

m4ximuel:
@WDaan I think I know what it feels like. So, how about using min-height and max-height at the same time so that the excess always uses the scroll bar? Like qbittorrent's configuration window.

WDaan:
I think that would be the best solution :slightly_smiling_face:

m4ximuel:
Untitled-1

I found the cause. coding...

WDaan:
I think we're ready to merge, no?

m4ximuel:
@WDaan All the tests for applying the new icon have been completed. Shall we apply it?

WDaan:
Yes, but in a seperate PR!
Let's merge this one & start fresh

LukePrior:
This looks exciting can't wait for the new icon changes to be merged and 0.9.0 to get released.

m4ximuel:
@LukePrior , @WDaan  Soon :)

IevgenSobko:
Impatiently waiting for a new release.
I guess it is going to skyrocket user experience.

WDaan:
@m4ximuel I think we could do a release already, right?

m4ximuel:
we're ready!

IevgenSobko:
It just an awesome release.
Totally different feel on an Android device.
Rounded corners in the context menu, autofocusing for search text, working scroll in modals in a mobile browser.
And it is just the tip of the iceberg that hidden underdense this release refactoring.

WDaan:
Thanks for the kind feedback! I'm really glad you like it. :)

Major thanks to @m4ximuel! He did a lot of great work on this one

PR #208, opened 2021-04-10
Change the context menu mechanism

m4ximuel:
# Features
- Right-click menu available for iPhone and most devices. Replaced by Long Touch
# Fixes
- Removed 'context-menu' package that is no longer available


WDaan:
This context menu is insane!! I really like it, makes it way easier to use.
On mobile the transition speed should be a little faster I think, it feels a tad sluggish

m4ximuel:
@WDaan Standard milliseconds for user response. Would it be okay to adjust it to 500ms->400ms?
Also, there is a problem with opening the sub-menu on mobile, so I'm in the process of adjusting.

WDaan:
Sure, maybe even 300ms... as long as it feels responsive.

What seems to be the issue?? I tested it and for me it worked perfectly fine :)

m4ximuel:
@WDaan 
When the touch mode is detected, the menu is not closed when the sub-menu is clicked.
:open-on-hover="!touchmode"
Greater compatibility.

m4ximuel:
@300ms, ok

m4ximuel:
merge to #207

PR #212, opened 2021-04-13
chore: use process.env for version

jef:
# chore

Use `process.env` for version instead of importing `package.json`. Completely unnecessary, but something I learned recently and could be better practice?

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -16,7 +16,6 @@
 <script>
 import { mapState, mapGetters } from 'vuex'
 import Navbar from '@/components/Navbar/Navbar.vue'
-import { version } from '../package.json'
 import qbit from '@/services/qbit'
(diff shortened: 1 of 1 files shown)


WDaan:
I didn't know you could do that! Very neat! 😮😮

jef:
It seems that building the files doesn't evaluate the version, not even during runtime, which is strange. I'm going to either revert this or figure out why this doesn't work. Will follow up in a PR.

WDaan:
The webui doesn't really 'run' like a backend service, only in when developping. Once 'compiled', the files are static. That's probably why it doesn't work. Still weird that it doesn't pick those variables up when building

jef:
Fixed!

Issue #60, opened 2020-10-26
Add Keyboard navigation in Torrent Details - Content tab


EduCampi:
Example:
Torrent with multiple files where we don't want to download all of them.
Right now, one has to scroll down and click one at the time, it would be way quicker if we could navigate with down/up arrows and use the spacebar to select/deselect.

Being able to hold shift to select from one file to another one would also be great.

Thanks,

WDaan:
With the current implementation, I don't think either of those solutions is possible.

There is an open issue about it in the framework though, so it might get solved: https://github.com/vuetifyjs/vuetify/issues/10796

m4ximuel:
I'm working on for this.

WDaan:
Really? It's a tough one to solve, the framework itself doesnt even have a proper solution.

Maybe use the Google material datatable instead of the vuetify one?

Good luck! Let me know if you need any help

m4ximuel:
@WDaan #216

PR #216, opened 2021-04-15
Control content tab use keyboard

m4ximuel:
# Control content tab use keyboard [feat]
- Move the list index with the up and down arrow keys
- Expand/collapse the list with the left and right arrow keys
- Toggle exclude/include files with space and enter keys

# PR Checklist
- [x] All Unit tests pass


m4ximuel:
@WDaan I thought I made a PR from the 'master', but there is an unexpected history. Shall I make a new PR?

WDaan:
No it's fine :) it only shows changes related to this PR. That's what matters.

I will test this out tomorrow! It looks impressive.

m4ximuel:
I found a bug in the treeview.
Items with the same file name are not selected at the same time, but items with the same folder name are selected at the same time. Please solve this problem.

m4ximuel:
@WDaan Functionally, it is complete, but please review whether that part of the bug is a fixable problem. I can't modify the treeview structure because it can't be interpreted.

WDaan:
What do you mean by "it can't be interpreted"?

m4ximuel:
> What do you mean by "it can't be interpreted"?

I don't know how to edit tree data.

WDaan:
The 'tree' is generated with the treeify function. Once it's generated and passed into component, you cant do much with it anymore.

That's why I would like to get rid of the v-treeview component & look for another solution. One where we have more control over what happens

m4ximuel:
@WDaan hmm... Could you fix the first bug I mentioned before looking for a another solution?

> Items with the same file name are not selected at the same time, but items with the same folder name are selected at the same time.

It's out of my abilities. I can hardly understand the treeview.
Once you fix this, #60 can be closed satisfactorily.
Something... It seems to be possible to use the difference between the data included in the 'file' and the 'folder', but...

WDaan:
Closed due to inactivity

PR #218, opened 2021-04-18
Fix Global Speed Title

matusaurio:
Fix for global speed title [fix]
Download and upload speed were exchanged.


WDaan:
lol, what a stupid mistake of me. Great catch! thx 🙂

PR #222, opened 2021-04-18
fix: application version getter

jef:
# fix: application version getter

Currently, when webpack bundles the Vue application, it no longer has access to `process.env`. I've updated the webpack plugins to include this in the build and the application now has access to the value.

Ref: https://github.com/WDaan/VueTorrent/pull/212

I tested putting via `npm run serve` as well as `npm run build` and using it like we would in production. In both cases, seemed to be working fine!

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -100,10 +100,8 @@ module.exports = {
   configureWebpack: {
     devtool: 'source-map',
     plugins: [
-      new webpack.DefinePlugin({
-        APPLICATION_VERSION: JSON.stringify(
-          require('./package.json').version
-        )
+      new webpack.EnvironmentPlugin({
+        APPLICATION_VERSION: process.env['npm_package_version']
       })
     ]
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "vuetorrent",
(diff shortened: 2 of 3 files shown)


vue.config.js
```diff
@@ -100,10 +100,8 @@ module.exports = {
   configureWebpack: {
     devtool: 'source-map',
     plugins: [
-      new webpack.DefinePlugin({
-        APPLICATION_VERSION: JSON.stringify(
-          require('./package.json').version
-        )
```
jef, line 106: This wasn't in use, so I decided to have it be in use! :grin:

WDaan:
aha, so 'APPLICATION_VERSION' is known by default when building with webpack?

jef:
Yep! The `EnvironmentPlugin` specifies that this is a `process.env` scoped variable. If we use `DefinePlugin`, I think it's on the global scope. Not sure how that's referenced, but it seemed to me `EnvironmentPlugin` was shorthand for

```javascript
new webpack.DefinePlugin({
  'process.env': { 'APPLICATION_VERSION': process.env['npm_package_version'] }
})
```

WDaan:
Awesome! thx for this fix 🙂

jef:
Thanks for the quick turn around!

Issue #223, opened 2021-04-18
[Feature] new speed graph design without apexcharts.js


m4ximuel:
![graph](https://user-images.githubusercontent.com/54457152/115154233-c4213480-a0b4-11eb-92cd-b74d453a7f41.gif)

need change only SpeedGraph.vue.
It only use svg and javascript.
It has a positive impact on device performance.

WDaan:
I think the current graph looks way sexier. How much of a difference does it make? Maybe it becomes a user setting? Or can we tweak the visuals some more?

m4ximuel:
@WDaan Let’s fix it a little more beautifully. It's made with an obsession that likes simplicity.

WDaan:
Yes, great! :) 

I would like to see some code for it, like how easy it is to use/program.

Does it need dependencies? Because getting rid of Apex, would be nice.

m4ximuel:
anything no need.

m4ximuel:
![graph](https://user-images.githubusercontent.com/54457152/115199262-fde25180-a12d-11eb-9988-b6355b7e9064.gif)
@WDaan more pretty version

m4ximuel:
![graph](https://user-images.githubusercontent.com/54457152/115212044-173dca80-a13b-11eb-8247-6ee3c15ee516.gif)
@WDaan "ANIMATION"

WDaan:
This keeps getting better & better 😮😮

m4ximuel:
@WDaan When you have the time to spare, experiment with the latest source code for #224.
It also goes very well with VueTorrent's existing primary colors.

WDaan:
I will! I'm doing a full-time internship atm, so it can get quite busy 😅

m4ximuel:
#233

PR #224, opened 2021-04-18
Speed graph sample

m4ximuel:
# Speed graph sample [feat]
New design sample not using apex charts


m4ximuel:
I'm working for Optimizing performance

m4ximuel:
#223

Issue #225, opened 2021-04-18
Getting 'Unacceptable file type, only regular file is allowed' after a while


toniR15:
Hi, I'm running qbittorrent-nox 4.1.5 on a Raspberry pi 4 ( Raspbian 10) and I can download VueTorrent and apply the alterate webUI without problems and for some time it works fine. But after some time (about one hour more or less) with the system at idle, when I try to access it again the webUI, I get a blank screen with 'Unacceptable file type, only regular file is allowed' message.

After this happens the only way for me to solve it is to edit qbittorrent.conf and remove the flag for AlteranteWebUI to false, restart qbittorrent and then it can load again on the default web interface.

Any ideas of what could be going wrong and how I can solve it? Thanks!

m4ximuel:
It's possible that the storage is unmounted. This message is a message that occurs when qbittorrent cannot find the correct webui file.

jak3-taylor:
I'm running the same setup and mine doesn't even work for an hour, I only get that message.

FYI you can instantly get back to the old UI by adding this to the end of your url:
/api/v2/app/setPreferences?json=%7B%22alternative_webui_enabled%22:false%7D

rany2:
I've had this issue as well. I eventually gave up and just used Nginx reverse proxy instead

m4ximuel:
@toniR15  @rany2 I don't think the message was triggered by 'VueTorrent'.
First check this:
- The web ui file location must be the parent folder of the 'public' folder.
- Only 'public' folder should exist in web ui file location, and components of 'VueTorrent' such as 'index.html' should be included in 'public' folder.

rany2:
@m4ximuel Ah, so the WebUI path should be `/path/to/release` and not `/path/to/release/public`. My bad

m4ximuel:
@rany2 yes, you right. please try it and tell me the result.

rany2:
it works now, thanks!

toniR15:
> It's possible that the storage is unmounted. This message is a message that occurs when qbittorrent cannot find the correct webui file.

I really doubt that's the case since the VueTorrent folder is inside the home folder and it can still be accessed via ssh


> @toniR15 @rany2 I don't think the message was triggered by 'VueTorrent'.
> First check this:
> 
> * The web ui file location must be the parent folder of the 'public' folder.
> * Only 'public' folder should exist in web ui file location, and components of 'VueTorrent' such as 'index.html' should be included in 'public' folder.

[...]

PR #227, opened 2021-04-19
Establish Service WalkerRules

m4ximuel:
# Establish Service Walker Rules [fix]
Follow the Service Walker's cache policy. It now works in offline mode if PWA is not received.

# PR Checklist
- [ ] All Unit tests pass


m4ximuel:
#225

WDaan:
I'm not sure I fully understand what's going on here?

Is this THE long awaited PWA cache fix? 😝

m4ximuel:
@WDaan  If the service walkers are not same, it override the existing cache by new.
This operation may require closing and opening all VueTorrent pages 1-2 times depending on the environment in order to update, but this is the standard for service workers. without the work, this service worker cannot be installed as a PWA after August 2021.

PR #231, opened 2021-04-19
Modify torrent drag mode range and modify behavior

m4ximuel:
# Modify torrent drag mode range and modify behavior [feat/fix]
- The mode of adding a torrent by dragging it to the dashboard is now expanded throughout the document area.
- 'AddModal' will now open immediately without transition when Torrent is dragged.

# PR Checklist
- [x] All Unit tests pass


m4ximuel:
The feature to add torrent files by drag is included in the currently released version, but I don't think you noticed it.

PR #233, opened 2021-04-20
New Graphs

m4ximuel:
# new graphs [feat/perf]
- Change drawer graph design 
- add graphs to torrent cards in the dashboard
- 'Apex charts' has been removed.

# PR Checklist
- [x] All Unit tests pass
- [ ] I've removed all commented code


WDaan:
Amazing! 
Maybe put another screenshot in here, so it's easier to reference back to this in the future

Can we turn this new graph into an internal component? So it can be used for both use cases without duplicating code?

m4ximuel:
redesign
@WDaan Roles may be different in the future, so I want to keep them for now. For example, a drawer graph may have a prettier design, and the remaining capacity may be displayed as a graph.

WDaan:
Fair point!

For the SpeedGraph, I think it can stay as is. Because all the code is used to make a graph.
For the torrent graph, I would still make a different 'graph' component so the Torrent.vue component is not bloated by Graph code.

what do you think?

m4ximuel:
@WDaan I understood what it meant. I am not sure how to mixin this source code. Could you do it for this code?

WDaan:
Sure! Not sure when I will be able to do this though 😅 

I would probably make a Core/BaseGraph component with all shared stuff & then perhaps extend those where they're being used

m4ximuel:
How about saving the download/upload speed history to torrent data in the store data?

WDaan:
Hmmm I don't know, than we definetly should limit it to those that are shown on the dashboard. Now that happens automatically. 

Other than that, there's not much difference. Perhaps a little cleaner code.

m4ximuel:
@WDaan I've been using it for a few days, but it may be annoying or regretful that the download/upload speed data is initialized every time the page number is changed. How about store speed data even if the svg operation stops?

WDaan:
Yes, you make a valid point. That would definetly solve that issue.
It would also make the components code a lot cleaner, as we can just format/parse the data where we allready retrieve it... in the getMainData()

WDaan:
I'm trying to get all the new chart code into one or 2 customizable components so they can be used as a library & are easy to maintain. But this has proven a little more difficult than expected. Also the code to generate a chart is a little hard to read/understand, so I wanted to improve that as well.

I will try to finish this soon

WDaan:
Closing because it's been open for so long and I don't think anyone is going to be able to maintain this. The code seems too complex.

Shame because it looked awesome.

PR #238, opened 2021-04-22
New select mode color

m4ximuel:
# New select mode color [feat]
new select mode color for torrent lists

# PR Checklist
- [x] All Unit tests pass


WDaan:
If all select colors are the same as there status color I would take an easier approach 🙂 
In vuetify.js all colors are importerd as variables, so a simple ` :class="isSelected ?  torrent.state.toLowerCase() : ''"`
would work just the same, no extra css needed.

m4ximuel:
@WDaan Remove added styles and add tricks. And, instead of 'MobileCard' and 'DesktopCard', 'Torrent' now in charge of the side border style.

WDaan:
Very neat! 🚀

PR #255, opened 2021-05-02
Show connection status at the bottom of Navbar

aalyokhin:
# Connection status icon [feat]

Hey, thanks for a great project! Some time ago I was thinking about building alternative UI using the same stack, but when I found yours it was clear enough that there is no sense doing that.

Recently I was doing some network setup changes and found this little feature to be really missing - showing if connection port is open. To check that I had to use default web UI for some time, which, you know, is not the best experience :smile:. So here is my 5 cents and hopefully more to come.

4 icons will be shown for 4 states: _connected, disconnected, firewalled_ and when there is no status data, with a tooltip. Hopefully the placement is ok.
Please find screenshots below (all states on desktop + mobile view):

 

# PR Checklist
- [x] All Unit tests pass

### DIFF
--- a/src/components/Navbar/ConnectionStatus.vue
+++ b/src/components/Navbar/ConnectionStatus.vue
@@ -1,0 +1,40 @@
+<template>
+  <v-tooltip top>
+    <template #activator="{ on }">
+      <div class="d-flex justify-center fill-height">
+        <v-icon class="white--text" v-on="on">
+          {{ currentIcon }}
+        </v-icon>
+      </div>
+    </template>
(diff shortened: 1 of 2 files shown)


WDaan:
Hi! First of all, thank you for contributing! It's always nice seeing people collaborate on a project. 🙂 

This PR looks great, it's a nice feature to have. Love the implementation as well, it's almost like I've written it myself 😛 

Also, I really really appreciate the detailed analysis + screenshots... people tend to overlook those.

Issue #266, opened 2021-06-09
Login prompt even with AuthSubnetWhitelistEnabled=true


letienne:
Hello,

Thank you for the great work.

I tried bypassing authentication (done at the reverse proxy level) by enabling AuthSubnetWhitelistEnabled=true.

It "half works" as I'm no longer required to enter something in the login form, but I'm presented with a login prompt anyway, that I can simply dismiss by refreshing the page.

Is there any way to skip that unnecessary reload? Maybe by not redirecting to /#/login?redirect=%2F if the AuthSubnetWhitelistEnabled and AuthSubnetWhitelist are valid?

Thanks !

Issue #152, opened 2021-02-14
Login Dialog when configured to allow login from localhost


mrhotio:
If you configure qBittorrent to allow passwordless login from for example localhost, by default qBittorrent will not ask for a login prompt, with VueTorrent however you'll get a login dialog which you can just click without providing credentials. Maybe it would be nicer to just hide it altogheter.

binhex:
i too would love to see an option to turn off authentication prompts when set to passwordless (useful for reverse proxy configurations).

PR #310, opened 2021-10-24
feat: add disable authentication option

rare-magma:
# feat: Add disable authentication option
Adds a toggle option to disable the authentication by making the isAuthenticated function return true if the option is set in the UI and the user is already authenticated.
Solves #152 and #266 

# PR Checklist
- [x] All Unit tests pass


rare-magma:
I've updated the PR. 
I've added a check if it's forbidden to access the /app/version API endpoint before serving the login page. This way VueTorrent avoids asking for login when it's not needed.

WDaan:
Looks like a good solution!
I will try to test it out tomorrow 🙂

PR #373, opened 2022-02-09
Prefix enviromental variables according to Vue Cli documentation

chepe263:
# My Title [feat/fix/perf/chore]
In the file `vue.config.js` there are enviromental variables `QBITTORRENT_PORT`,  `VUETORRENT_PORT`, `QBITTORRENT_TARGET` are not picked up by Vue CLI unless they have a prefix `VUE_APP_`.

This PR changes the name of enviromental variables so they can be used when you build or serve the app.

Now it would be possible to use a dotenv file for those enviromental variables.
# PR Checklist
- [*] I've started from master
- [*] I've only committed changes related to this PR
- [*] All Unit tests pass
- [*] I've removed all commented code
- [*] I've removed all unneeded console.log statements

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -2,9 +2,9 @@ const webpack = require('webpack')
 
 const iconVersion = 22 //fingerprint
 
-const qBittorrentPort = process.env['QBITTORRENT_PORT'] ?? 8080
-const vueTorrentPort = process.env['VUETORRENT_PORT'] ?? 8000
-const proxyTarget = process.env['QBITTORRENT_TARGET'] ?? 'http://localhost'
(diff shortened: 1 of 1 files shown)


WDaan:
I think using `VUE_APP_xxx` makes them available in the frontend code. 
These envs are only used for proxying request during development & nowhere in frontend IIRC.

I don't see a real benefit to using a .env file for this usecase? 
It's only a few variables you rarely change

chepe263:
Exactly, but it is useful when you already have a qbittorrent server. Besides, the current enviromental variables are unusable. This new way you can use an existing server and you don't have to modify the code.

PR #385, opened 2022-03-04
perf: add Spanish Language

roky146:
Saw that this awesome WebUI doesn't have Spanis traslation! so i added it.

Sorry for my bad english :)


WDaan:
Great! Thank you :)

PR #406, opened 2022-04-16
feat: add support for setting global speed limits

giacomocerquone:
# Allowing the control of global speed limits [feat]

First of all, great project and a very clean code. Never used Vue before and I implemented this in less than an hour.
I tried to maximize the reuse in this pr that, as stated in the title, allows users to set global speed limits (both upload and download).
It reuses the SpeedLimitModal component that appears after clicking over the download and upload card speeds in the right drawer of the webpage.

Here is a screencast: https://watch.screencastify.com/v/a09SoF3H2Bt9flO7N1UI
(After settings the global limits you might observe some spikes in the reading of the speed data but the limits are correctly applied)

Closes #339
Enjoy 😃 

UPDATE: 

In a future update, we can also consider adding this stuff inside the settings

# PR Checklist
- [x] I've started from master
- [x] I've only committed changes related to this PR
- [x] All Unit tests pass
- [x] I've removed all commented code
- [x] I've removed all unneeded console.log statements

PR #408, opened 2022-04-16
Perf/global speed

WDaan:
## Description

Same as #406 but small changes for enhanced readability


src/services/qbit.js
```diff
@@ -227,10 +227,12 @@ class Qbit {
 
   getGlobalDownloadLimit() {
     return this.axios.get('/transfer/downloadLimit')
+      .then(res => res.data)
```
giacomocerquone, line 230: I've seen you using await here and there and to uniform it this IMHO should be rewritten like:

```
 async getGlobalDownloadLimit() {
   const {data} = await this.axios.get('/transfer/downloadLimit')
   return data
```

WDaan: Sure, I don't really have that big of a preference.... I've used both in that class.

PR #431, opened 2022-06-05
Fixed lowercase downloading in status selector

IevgenSobko:
fix: lowercase downloading in status selector


IevgenSobko:
Going to create one more with few fixes of casing

PR #432, opened 2022-06-05
fix: non-capital words

IevgenSobko:
fixed:
 - Status selector Downloading item case
 - torrents grid Upload, Downloaded, Uploaded label case
 - add torrent category selector title case
 - context menu set limit Upload case
 - context menu copy Hash misspelling


stillie:
Came here to make the exact same changes :)

Issue #311, opened 2021-10-26
[Feature] RSS Support (Hear me out!)


Clickbaitcake:
Before this issue is quickly closed please hear me out! I have read the other issues on this topic and can see no one really pressed good reason other than yourself, @WDaan if you wont use this feature then of course its fair enough that you wont develop it but id like to add my reason to the pile to see if I can convicne you...

I do not use Radarr or Sonarr because for me they are overkill. I dont use their features and dont really like having to maintain multiple services, I just have a box with lots of storage and your wonderful torrent client. For simplicity people like me might want to use https://showrss.info. Whenever a show is on it gets dropped into the RSS feed from ShowRSS and automatically downloads. Its great because the user does not have to worry about installing and maintaining Sonarr, Radarr and of course Jackett. This leaves the user with a nice simple server which focuses on what it does best, downloading torrents. I found 6 other requests for this issue with 12 thumbs up which means a rough total of 19 people (including myself) would like to see this feature!

If I had the know how I would build this and submit a pull request but I just dont. Although @WDaan  you owe us nothing, we would all really appricate this feature. Im sure some folks would even donate a coffe or two... I know I will.
Paging other supporters of this idea: @NilSilva @Zonimi @Azraelle @kbipinkumar

In any case, thank you for your development work so far already, this is a great project!

will-blk:
It's a feature I also use a lot. That will be my first attempt with VUE but I will try it out. 

My goal will be:

- Enable/Disable rss auto download
- Add feed
- Add rule

I also have a full-time job so whatever comes next will depende on how long I can get it done x)

How that sounds @WDaan ?

WDaan:
@will-blk 

Looking forward to your PR. Let me know if you need help.

will-blk:
Actually, I do. I'm trying to use addFeed api but couldn't get it right. Only returns 400

The request: `this.execute('post', '/rss/addFeed', { url: feed.url })` where feed is: `{ url: 'whatever'}`

The references: 
- [github](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#add-feed)
- [readthedocs](https://qbittorrent-api.readthedocs.io/en/latest/apidoc/rss.html)

I was able to list and delete feeds and rules already but couldn't figure those creations out, tried both params options. I'm looking the right doc?

will-blk:
Finally, got that creation working!

WDaan:
What changed??

[...]

PR #447, opened 2022-06-26
Basic rss interface

will-blk:
# Rss basic interface
- Enable/Disable background processing option
- Enable/Disable autodownload articles option
- Create/Delete feeds
- Create/Delete rules (Only supports "must contain" input and affected feed)

# PR Checklist
- [x] All Unit tests pass

Issue #311 resolves


src/lang/en.js
```diff
@@ -191,6 +203,29 @@ const locale = {
         whenRatioReaches: 'When ratio reaches',
         whenSeedingTimeReaches: 'When seeding time reaches'
       },
+      pageRss: {
+        tabName: {
+          general: 'Geral',
```
WDaan, line 208: this seems like a typo :p

WDaan: Also, I think you need to add 'feed' at the top here in the 'General' section... at the top of the file

src/services/qbit.js
```diff
@@ -154,6 +154,42 @@ class Qbit {
...
+
+  createRule(ruleName, defs) {
+    return this.execute('post', '/rss/setRule', {
+      ruleName: ruleName,
+      ruleDef: JSON.stringify(defs)
+    })
+  }
+
+  getFeeds() {
+    return this.axios.get('/rss/items').then(res => res.data)
```
WDaan, line 174: I'd sanitize the data here 

```
    return this.axios.get('/rss/items')
      .then(res => res.data)
      .then(data => 
        Object.entries(data).map(feed => {
          return { name: feed[0], ...feed[1] }
        })
      )
```


Same for rules 
```
  getRules() {
    return this.axios.get('/rss/rules')
      .then(res => res.data)
      .then(data => Object.entries(data).map(rule => {
        return { name: rule[0], ...rule[1] }
      }))
  }
```

src/store/mutations.js
```diff
@@ -80,6 +80,18 @@ export default {
     state.sort_options.tracker = tracker
   },
   FETCH_CATEGORIES: async state => state.categories = Object.values(await (qbit.getCategories())),
+  FETCH_FEEDS: async state => {
+    const feeds = await qbit.getFeeds()
+    state.feeds = Object.entries(feeds).map(feed => { 
+      return { name: feed[0], ...feed[1] }
+    })
+  },
+  FETCH_RULES: async state => {
```
WDaan, line 89: we can keep this clean if we sanitize/format the data when doing the api call

```
  FETCH_FEEDS: async state => state.rss.feeds = await qbit.getFeeds(),
  FETCH_RULES: async state => state.rss.rules = await qbit.getRules(),
```

WDaan:
Awesome work! Seems to be working perfectly :) 

I've commented some minor changes, I can fix them myself if you want.

Create/Delete on Rules is okay... but we should support the full form in the future
Also the feed multi-select part in the create rule will not be ideal for people with a lot of feeds, but I guess that's also a later improvement.


Great work so far :) thanks

src/lang/en.js
will-blk: A mix of english and portuguese hahaha Will fix it

will-blk:
Thanks :smile: I will make the fixes soon, haven't seen the store index file before. 

As an improvement idea, I would also move RSS from the settings and make a tab elsewhere. I think It will become too small after adding the feed preview...

will-blk:
Check it out!

PR #506, opened 2022-10-03
feat: Links in torrent details view are now clickable

tadasvosylius:
I have added filter to make _"Torrent Details"_ > _"Info"_ links clickable. Filter finds URL and replace it to <a> tag, so it becomes clickable. I have added this feature only to **tracker**, **created by** and **comment** fields, because only those fields can contain URL. This filter **will not** modify URLs which start without www/http/https (i.e. _something.com_) to prevent accidental replacement when sentence ends and new one starts without a space (i.e. _something cool.To replace_)

Also I'm participating in hacktoberfest, so it would help me alot if you can add "hacktoberfest-accepted" label to this PR :)

Related issue: https://github.com/WDaan/VueTorrent/issues/477

Pics:
Screenshot from 2022-10-03 22-37-48

# PR Checklist
- [x] All Unit tests pass

### DIFF
--- a/src/components/TorrentDetail/Tabs/Info.vue
+++ b/src/components/TorrentDetail/Tabs/Info.vue
@@ -139,25 +139,19 @@
           <td :class="commonStyle">
             {{ $t('modals.detail.pageInfo.trackers') }}
           </td>
-          <td>
-            {{ torrent.tracker }}
-          </td>
+          <td v-html="$options.filters.transformUrlToHref(torrent.tracker)" />
         </tr>
         <tr v-if="createdBy">
           <td :class="commonStyle">
             {{ $t('modals.detail.pageInfo.createdBy') }}
           </td>
-          <td>
-            {{ createdBy }}
-          </td>
+          <td v-html="$options.filters.transformUrlToHref(createdBy)" />
         </tr>
         <tr v-if="comment">
           <td :class="commonStyle">
             {{ $t('torrent.comments') | titleCase }}
           </td>
-          <td>
-            {{ comment }}
-          </td>
+          <td v-html="$options.filters.transformUrlToHref(comment)" />
         </tr>
 
         <tr>
--- a/src/filters.js
+++ b/src/filters.js
@@ -131,3 +131,11 @@ export function limitToValue(value) {
 }
 
 Vue.filter('limitToValue', limitToValue)
+
+export function transformUrlToHref(string) {
+  const expression = /(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/gi
+
+  return string.replace(expression, '<a href="$1" target="_blank">$1</a>')
+}
+
+Vue.filter('transformUrlToHref', transformUrlToHref)
--- a/tests/unit/filters.spec.js
+++ b/tests/unit/filters.spec.js
@@ -1,8 +1,28 @@
-import { titleCase } from '../../src/filters'
+import { titleCase, transformUrlToHref } from '../../src/filters'
 
 describe('Filters', () => {
   it('titleCase', () => {
     expect(titleCase('test')).toEqual('Test')
     expect(titleCase('hello there')).toEqual('Hello There')
+  })
+
+  describe('transformUrlToHref()', () => {
+    it('should not transform anything', () => {
+      expect(transformUrlToHref('test')).toEqual('test')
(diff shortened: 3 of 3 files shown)


WDaan:
Awesome, thanks for providing tests as well :)

WDaan:
I left 1 comment about the filter

tadasvosylius:
> I left 1 comment about the filter

Hmm, where I can see it? Because in this conversation, I can see only this one:

> Awesome, thanks for providing tests as well :)

WDaan:
@tadasvosylius It's in the review :) (My apologies for the delay)

Screenshot 2022-10-05 at 10 18 24

tadasvosylius:
> @tadasvosylius It's in the review :) (My apologies for the delay)
> 
> Screenshot 2022-10-05 at 10 18 24

Ahh, sorry for the rush :) I will wait until review is done, thank you for information :)

src/components/TorrentDetail/Tabs/Info.vue
```diff
@@ -139,25 +139,19 @@
...
-            {{ createdBy }}
-          </td>
+          <td v-html="$options.filters.transformUrlToHref(createdBy)" />
         </tr>
         <tr v-if="comment">
           <td :class="commonStyle">
             {{ $t('torrent.comments') | titleCase }}
           </td>
-          <td>
-            {{ comment }}
```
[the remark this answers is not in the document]
WDaan: Seems like I forgot to press save 😭 
Other than that everything looks great :)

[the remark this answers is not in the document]
tadasvosylius: I totally agree with you that it looks much cleaner. But the problem is that using this way, filter always returns _string_, and shows all html elements as string. So output from filter would look like "<a href='smth.else'>smth.else</a>". That's why I have used v-html directive - so it will insert content as plain HTML :)

src/components/TorrentDetail/Tabs/Info.vue
```diff
@@ -139,25 +139,19 @@
...
-          </td>
+          <td v-html="$options.filters.transformUrlToHref(comment)" />
```
V-ed, line 154: Isn't this a very easy to achieve XSS vulnerability? Since comments on torrents could be anything

(and even to an extent the trackers, although I am pretty sure those should generally be fine)

V-ed: Frankly speaking I don't want to update to v0.19.0 until this is addressed :/

https://vuejs.org/guide/essentials/template-syntax.html#raw-html

image

tadasvosylius: Very good catch, my bad. I will try to solve this problem later in the evening. I think, that best solution is to sanitise string and only then transform url to anchor tag, what do you think @V-ed @WDaan?

WDaan: I've temporarily disabled the feature. 

I was thinking about iterating the string and actually rendering the parts that match as an <a> with a dynamic component, instead of string replace v-html.

But than you need a way to identify a part as match or non match

tadasvosylius: Okay, I will try to implement this logic

V-ed: > I've temporarily disabled the feature. 

Thank you for the quick actions!

It's a bummer though, clickable links in the fields is also something I see myself wanting soon haha, I'll keep an eye out on how one would achieve this with dynamic components in vue :)

turikhay: I've also discovered this yesterday (check your email for silly PoC @WDaan :) )

This change does not, however, allow  `<script>` tags to be included in DOM.

WDaan: Yes, I saw the email and immediately reverted. Thanks for notifying.

Script tags not working is something. But if we want this feature, we should think of a way to properly implement it.

V-ed: > This change does not, however, allow `<script>` tags to be executed.

What do you mean @turikhay ? I haven't tested it and I am not the most familiar with Vue but I'm fairly certain (please correct me if I am wrong here) that the `v-html` doesn't do any stripping and the `transformUrlToHref` method implemented only searches for links, so I am not sure what do you mean by "script tags won't be executed" - as soon as they will be "rendered" they will be run no?

Besides, you can do XSS even without direct script tags (``) but that's another story.

turikhay: @V-ed you're right, Vue doesn't filter anything when `v-html` is used.

Last time I checked plain `<script>` wasn't executed. It just didn't.

I assumed it was because server sent appropriate CSP headers to block that kind of things. But it wasn't the case: qbittorent [doesn't even send them](https://github.com/qbittorrent/qBittorrent/blob/4f0990dc297c40ba3047a539d3bce10803d67c78/src/webui/webapplication.cpp#L372) when third-party WebUI is used.

However, I can confirm that `` will trigger XSS. Thanks for the tip.

Issue #483, opened 2022-08-16
Missing options available in standard WebUI


Bust3r14:
![Screenshot 2022-08-16 144817](https://user-images.githubusercontent.com/42153444/184956914-39cd4111-7b6e-4cb5-99e4-7d56118a95dd.png)
![Screenshot 2022-08-16 144924](https://user-images.githubusercontent.com/42153444/184956918-28f7776f-2d7e-47c7-98f4-8f32b61e406c.png)
The VueTorrent UI does not have options in it's settings for what to do with torrents once they reach certain criteria; the criteria options exist, but the resulting actions (pause, remove, superseed, etc) do not exist.

WDaan:
You're right this is more of a general issues, settings options are being added as we go or when they're requested.

giacomocerquone:
I'm going to work on this 😃

PR #523, opened 2022-10-19
feat: add max_ratio_act handling

giacomocerquone:
# feat: add max_ratio_act handling

Implements #483 

image

Wanted to open this as a draft since it's almost finished but also because I had to reverse engineer the thing a little bit here because [the following documentation](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#get-application-preferences) doesn't show correctly all the possible values assignable to the max_ratio_act param.

I've inspected it from the graphical qbittorrent instance running on my mac that the values are indeed four and go from 0 to 3 and the order is a bit fucked up. Look at this screen I've made that maps the index of the options within the select to the actual values:

image

# PR Checklist
- [x] All Unit tests pass

### DIFF
--- a/src/components/Settings/Tabs/BitTorrent.vue
+++ b/src/components/Settings/Tabs/BitTorrent.vue
@@ -171,11 +171,13 @@
       </v-row>
     </v-list-item>
     <v-list-item>
-      <v-row>
-        <v-col offset="6">
+      <v-row dense>
+        <v-col>
           <v-subheader>
             {{ $t('then') }}
           </v-subheader>
+        </v-col>
+        <v-col>
           <v-select
             v-model="settings.max_ratio_act"
             class="mb-2"
--- a/src/components/Settings/Tabs/BitTorrent.vue
+++ b/src/components/Settings/Tabs/BitTorrent.vue
@@ -175,7 +175,7 @@
       <v-row>
         <v-col>
           <v-select
-            v-model="settings.seeding_limits_then"
+            v-model="settings.max_ratio_act"
             outlined
             :disabled="!settings.max_ratio_enabled && !settings.max_seeding_time_enabled"
             dense
@@ -199,11 +199,11 @@ export default {
     return {
       thenTypes: [
         {
-          value: 1,
+          value: 0,
           text: 'Pause torrent'
         },
         {
-          value: 2,
+          value: 1,
           text: 'Remove torrent'
         },
         {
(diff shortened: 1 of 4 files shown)


src/components/Settings/Tabs/BitTorrent.vue
```diff
@@ -178,6 +194,28 @@ import { SettingsTab, FullScreenModal } from '@/mixins'
...
 export default {
   name: 'BitTorrent',
-  mixins: [SettingsTab, FullScreenModal]
+  mixins: [SettingsTab, FullScreenModal],
+  data() {
+    return {
+      thenTypes: [
+        {
+          value: 0,
+          text: 'Pause torrent'
```
WDaan, line 203: Should this be keys for vue-i18 so we can easily translate them?

WDaan (commented):
Awesome! Thanks for picking this up!

2 suggestions:
- maybe make the options vue-i18n keys so people can easily add translation support
 - current layout looks a bit funky... maybe the drop-down should on the right side with the same width as the others? With the `then`-keyword on the left side

src/components/Settings/Tabs/BitTorrent.vue
giacomocerquone: Don't know... I followed the methodologies within the proxy settings in another screen. If you want I can add these within i18n

WDaan: The request will probably come sooner or later, so might as well just do it from the start

giacomocerquone:
it should be done I'm fighting with this though 
Schermata 2022-10-19 alle 22 21 36

Any advice?

WDaan:
Something like this is what I had in mind.

Screenshot 2022-10-20 at 09 10 30

Using the following snippet:
```vue
  <v-list-item>
      <v-row dense>
        <v-col>
          <v-subheader>
            {{ $t('then') }}
          </v-subheader>
        </v-col>
        <v-col>
          <v-select
            v-model="settings.max_ratio_act"
            class="mb-2"
            outlined
            :disabled="!settings.max_ratio_enabled && !settings.max_seeding_time_enabled"
            dense
            small-chips
            :items="thenTypes"
          />
        </v-col>
      </v-row>
   </v-list-item>
```

giacomocerquone:
done, also fixed tests 🙂

WDaan:
Awesome! Nice addition 👍

PR #551, opened 2022-11-23
fix: Unable to rename torrent

Larsluph:
# Unable to rename torrent [fix]

This PR fixes the `qbit.setTorrentName` using `GET` method instead of `POST`.

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/services/qbit.ts
+++ b/src/services/qbit.ts
@@ -125,12 +125,7 @@ export class QBitApi {
   }
 
   setTorrentName(hash, name) {
-    const params = {
-      hash,
-      name
-    }
-
-    return this.axios.get('/torrents/rename', { params })
+    return this.execute('post', '/torrents/rename', { hash, name })
   }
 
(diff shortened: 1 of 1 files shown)


src/services/qbit.ts
```diff
@@ -125,12 +125,10 @@ export class QBitApi {
...
 
   setTorrentName(hash, name) {
-    const params = {
-      hash,
-      name
-    }
-
-    return this.axios.get('/torrents/rename', { params })
+    return this.execute('post', '/torrents/rename', {
+      hash: hash,
```
WDaan, line 129: Just nitpicking here but, can be rewritten as just `hash`

WDaan (approved):
Awesome! Thanks for contributing ;)

src/services/qbit.ts
Larsluph: done!

PR #552, opened 2022-11-24
fix: Enabled category creation

Larsluph:
# Enabled category creation [fix]

Removed return statement that blocked category creation action

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/components/Modals/TagsAndCategories/CreateCategoryDialog.vue
+++ b/src/components/Modals/TagsAndCategories/CreateCategoryDialog.vue
@@ -72,8 +72,6 @@ export default {
     create() {
       this.$refs.categoryForm.validate()
 
-      return
-
       qbit.createCategory(this.category)
       this.cancel()
     },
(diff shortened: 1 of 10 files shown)


WDaan:
😬 good catch... lack of proper tests makes things like this slip through. 

Must've forgotten when improving the validation. 

Shouldn't it still early return when validation is falsy though? By removing the return it will probably still continue.

WDaan (approved):

Larsluph:
Now that you say it that may be a good thing to test, i'll check that!

Larsluph:
Okay so I checked that validation method and it's useless as-is. With that, I removed the length limitations which are not present in the qBittorrent WebUI API.

WDaan:
Awesome! 

The 2 translations can be cleaned up as well.


Don't know why the limit was introduced in the first place.

Larsluph:
FYI, certain translation mappings are missing some keys: es, it, nl, pt-br, vi

WDaan:
I know, but I'm not able to translate those. So, depending on the community.

Larsluph:
Maybe you can add a segment on that in the readme ?

WDaan:
That's a good idea. 
I'm also trying to find a tool that can show which translations are missing in the other files, but haven't found exactly what I'm looking for just yet.

PR #573, opened 2022-12-29
Update and realign locales

Larsluph:
# Update and realign locales [feat]

Add previously untranslatable fields and realign all locales to make contributions easier.

# PR Checklist

- [x] All Unit tests pass


WDaan:
I guess this one can be rebased from master??

I was thinking about building a command that updates the translation files. But haven't found an existing solution just yet.
So a command/tool that looks at the en.js and uses that to populate non-existing fields of the other translation files.

Larsluph:
Something like that would be awesome but no idea how to put that in place 😅

src/store/index.js
```diff
@@ -73,44 +93,8 @@ export default new Vuex.Store({
...
-        { name: 'Seeds', active: true },
-        { name: 'Status', active: true },
-        { name: 'Ratio', active: true },
-        { name: 'Category', active: true },
-        { name: 'Tags', active: true },
-        { name: 'AddedOn', active: true },
-        { name: 'Availability', active: true },
-        { name: 'LastActivity', active: false }
-      ]
+      busyTorrentProperties: [...propertiesTemplate],
```
WDaan, line 96: nice cleanup, however now they have the same state. I think there were some differences between the 2. But doesn't matter that much...this is cleaner.

PR #599, opened 2023-01-08
Add globalSpeed and globalVolume computed values to torrent properties

Larsluph:
# Add globalSpeed and globalVolume computed values to torrent properties [feat]

Compute global values (sum of download and upload) for transfer volume and transfer speed.

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,27 +2,30 @@
 
 ## [1.1.0](https://github.com/WDaan/VueTorrent/compare/v1.0.1...v1.1.0) (2022-12-30)
 
-
 ### Features
 
-* Add Batch change location ([#574](https://github.com/WDaan/VueTorrent/issues/574)) ([060c3c3](https://github.com/WDaan/VueTorrent/commit/060c3c3a1c463e988014fe1ae52569921a68d545))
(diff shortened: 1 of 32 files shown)


WDaan:
Could you explain this feature a little bit more please? I'm not quite sure I follow

Larsluph:
The only "useful" diff in this Pr are these : https://github.com/WDaan/VueTorrent/pull/599/files#diff-df87f1db7ad372d7981cd494af44c6f1934a6eaf30f2fa19fee4334a3cd32407R49-R52

I just added a torrent property to be able to display and sort by the total bandwidth used by a single torrent (globalSpeed), I also added globalVolume because it's literally the same thing

WDaan:
Aha! I understand, neat 😎 

Ready to merge?

WDaan:
Thanks for checking 😃

Larsluph:
I need your advice @WDaan for the design of this fix
To disable sorting using the API, is it better to
- Create an array with the name of the properties to exclude
- Use a prefix (like `computedGlobalSpeed`) and disable it if the property name starts with `computed`

Knowing that I plan to improve the feature by adding a local sorting system where everything is handled on our end (which may allow for better sorting customisation)

WDaan:
I guess an array because it allows for greater flexibility later on?

Larsluph:
Everything should work now! 🚀

PR #627, opened 2023-01-27
perf: Rework desktop cards

Larsluph:
# Rework desktop cards [perf]

Add flex gap to Desktop Card to improve item separation lisibility and reworked margins.

# PR Checklist

- [x] All Unit tests pass


WDaan:
Still trying to convert the components? 😝

Seems like most things are working except the one test, probably some small mistake with the status constructor (expoter?)

Larsluph:
Yup I'm converting the components when I need to change things in them. I'm trying to fix that error but I really don't understand what's the issue 😅

WDaan:
As for the `helpers.js` you can convert it to `.ts` maybe into a seperate folder so you can split up the helpers.

```
export function getVersion() {
  if (isProduction()) {
    return 'import.meta.env.VITE_PACKAGE_VERSION'
  }

  return import.meta.env['VITE_PACKAGE_VERSION']
}

export function getBaseURL() {
  if (isProduction()) {
    return 'import.meta.env.BASE_URL'
  }

  return import.meta.env['BASE_URL']
}
```

These ones that are giving issues, we can keep in a `.js` for now, will solve later.

WDaan:
Looking into the failing test, not sure yet why it isn't working

Larsluph:
For the helpers, I still got the parse error when I'm separating things

WDaan:
replacing `this.$i18n.t` by `this.$t` seems to have made it better 😛

Larsluph:
Insane, seems like I will never understand JS sometimes 😂
Good job!

PR #659, opened 2023-02-16
feat: Add import / export settings button

Larsluph:
# Add import / export settings button [feat]

Add buttons to import / export settings configuration, UI/UX design might be reworked 😅

image

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/components/Settings/Tabs/VueTorrent/VGeneral.vue
+++ b/src/components/Settings/Tabs/VueTorrent/VGeneral.vue
@@ -207,8 +207,15 @@
         </v-col>
       </v-row>
     </v-list-item>
+    <v-list-item>
+      <v-textarea v-model="settingsField" />
+    </v-list-item>
+    <v-list-item class="remove-after justify-content-evenly">
+      <v-btn @click="importSettings">{{ $t('modals.settings.pageVueTorrent.pageGeneral.importSettings') }}</v-btn>
+      <v-btn @click="exportSettings">{{ $t('modals.settings.pageVueTorrent.pageGeneral.exportSettings') }}</v-btn>
+    </v-list-item>
     <v-list-item class="justify-center pb-2">
-      <v-btn @click="resetSettings"> {{ $t('modals.settings.pageVueTorrent.pageGeneral.resetSettings') }} </v-btn>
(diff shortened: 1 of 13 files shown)


WDaan:
This is neat! Why didn't I think of this?!

Any way we could do some JSON formatting/validation on pasting?? Might be a bit easier to understand/debug when importing settings

Larsluph:
I looked into it this afternoon, I only found something to validate that it's a JSON but not to validate a certain JSON structure, no idea if it can help to migrate this component to TS 🤔

src/components/Settings/Tabs/VueTorrent/VGeneral.vue
```diff
@@ -242,6 +252,20 @@ export default {
     async fetchQbitVersion() {
       this.Qbitversion = await Qbit.getAppVersion()
     },
+    importSettings() {
+      try {
+        const userState = JSON.parse(this.settingsField)
+      } catch (e) {
+        alert('Invalid JSON!')
```
WDaan, line 259: Would probably be better to replace this with something like 
`Vue.$toast.error(....)`

WDaan:
About the validation... I think we might be able to check it against something like `Partial<StoreState>` to enforce the schema... Or maybe use something like JsonSchema.
I'm not yet sure what the best approach would be for something like this.

Larsluph:
Yeah but about that I saw that TS is only check types at compile time, at runtime it's just plain JS so it can't type check in our case

WDaan:
Then perhaps something like JsonSchema, we could generate it from the Typescript type and validate against it. It would probably give some better errors about what's wrong with the provided settings upon invalid configuration.

Larsluph:
Just dropping this here to find it back easily
https://www.npmjs.com/package/jsonschema

Larsluph:
Okay so jsonschema works when I'm using it in a single file, while importing it using `require('jsonschema').validate`.

Problem is that when using `import` statements, the `validate` function doesn't work and throws an error: `resolve is not a function`
https://github.com/tdegrunt/jsonschema/blob/5e219db670bb8a57f5af932e21d51e2ccf44e679/lib/validator.js#L118

Larsluph:
I think I left the json schema check that fails in the PR, the feature might not be usable in the current state.

Issue #676, opened 2023-02-24
[Feature Request] show tracker domain in torrent card


flashlab:
It'll be very convenient to directly mark the tracker domain in the torrent dashboard, especially in the context of reseeding. 
Just like the blue tag in mock images below:
![desktop view](https://user-images.githubusercontent.com/4102853/221136339-6e6e91ea-4fc9-481e-a170-d8e1cfc35f1d.png)
↑↑↑desktop view↑↑↑
![mobile view](https://user-images.githubusercontent.com/4102853/221136362-c942c920-6aa8-456b-84a9-ae478553d25c.png)
↑↑↑mobile view↑↑↑
For example, if the torrent's tracker url is like "https://tracker.xxx.com/...", the marker canbe "xxx".
Currentlly, I have to double click the card and locate to the tracker tab to get such information for each torrents.

May this feature be achieved?

2fst4u:
I see this has been implemented in 1.4.0, but can I please request that before this closed that this is made optional? I use qbittorrent-manage which automatically tags all torrents based on tracker so now I have two labels indicating tracker. I only need one of them.

flashlab:
@2fst4u the tracker source tags will not show on desktop view by default, and you can change the behavior in settings->vuetorrent->dashboard. 
BUT as to mobile view, there is no switch to remove it.maybe we can align the settings to solve it.

2fst4u:
> @2fst4u the tracker source tags will not show on desktop view by default, and you can change the behavior in settings->vuetorrent->dashboard. 
> BUT as to mobile view, there is no switch to remove it.maybe we can align the settings to solve it.

Yea I almost exclusively use mobile so that will be why I'm seeing it all of a sudden. Aligning them would be great.

Larsluph:
Now that I think of it, it may be possible to just toggle which data we want to display on the card, without being able to reorder them like in the desktop card

PR #679, opened 2023-02-25
feat: Add tracker host in Dashboard

flashlab:
#  Add tracker host in Dashboard

This will be useful when we use qBit as PT client. Especially when we have many reseeded enties across seveal sites.
more details are in #676 .

# PR Checklist

- [x] All Unit tests pass


src/components/Torrent/DashboardItems/Tracker.vue
```diff
@@ -0,0 +1,40 @@
...
+
+export default defineComponent({
+  name: 'Tracker',
+  mixins: [TorrentDashboardItem],
+  props: {
+    torrent: Torrent
+  },
+  computed: {
+    trackerString() {
+      let trackerUrl = this.torrent.tracker.match(/:\/\/([^\/]+\.)?([^\/\.]+)\.[^\/\.:]+/i)
```
WDaan, line 25: Can we extract this matching, that's also used in the other function to a helper or something?

Perhaps write some unit tests for it to guarantee its functionality. 🤔

WDaan (approved):
Thanks for becoming a contributor! Cool feature :) 

Left some small remark that would improve maintainability.

src/components/Torrent/DashboardItems/Tracker.vue
flashlab: I'm not familiar with nodejs, maybe the second commit will be better ;)

WDaan:
Awesome! 😎

PR #715, opened 2023-03-18
fix: Ctrl key not being handled on Windows/Linux

Larsluph:
# Ctrl key not being handled on Windows/Linux [fix]

Missing brackets so condition was always evaluated to true cause function exists 😬 

# PR Checklist

- [x] All Unit tests pass


WDaan:
Lol good catch, stupid mistake

PR #756, opened 2023-04-09
Migrate to Vue 3 (complete rewrite)

LeaderbotX400:
Migrate to Vue 3


WDaan:
Seems like you pretty much started the project over and migrate components. I'm not that big a fan of that approach because  we'll have to do a tons of things over all over again.

However, when following the migration docs I've noticed that it's a bit harder to get everything working again than anticipated.
I've started off with something like this https://github.com/WDaan/VueTorrent/pull/758.

I'm a bit at a crossroads right now.

WDaan:
Seems like redoing is indeed the better path, unfortunately.

(Because it requires a lot of effort before we can merge to main and actually see some of the result/benefits)

But we can cleanup and redesign a lot more this way, which is nice.

PR #761, opened 2023-04-12
Feat: Update things to vue 3

LeaderbotX400:
Update SpeedCard and SpeedGraph components to look better / actually work
Replace plain axios with vueuse axios for better reactivity using refs


LeaderbotX400:
Added the option to override the auth system when in dev mode. it does not actually login and as such is unable to fetch data / change anything but I think it's useful to test the UI even if you don't have a dev server to test with

PR #763, opened 2023-04-13
chore: update chinese translate

Moius:
# Chore: update chinese translate

Update the chinese translate.(simple and traditional)

# PR Checklist

- [x] All Unit tests pass


VEWION:
I am sorry, but I believe the Traditional Chinese translation is not suitable. I have modified it based on the original web UI to minimize confusion for people switching from the original to VueTorrent. The word choices are based on usage in Taiwan, as is common in most software.

Moius:
> I am sorry, but I believe the Traditional Chinese translation is not suitable. I have modified it based on the original web UI to minimize confusion for people switching from the original to VueTorrent. The word choices are based on usage in Taiwan, as is common in most software.

I just updated the translation structure of Traditional Chinese and simply translated from Simplified Chinese to Traditional Chinese. I am from mainland China, so many translations may not be appropriate, you have a more suitable translation please translate and submit a new PR based on Taiwan's habits, thanks!

VEWION:
Also I have concerns about the translation in line 138, as mentioned in #594. I believe it would be better to leave it unchanged to avoid this awkward appearance.
螢幕擷取畫面 2023-04-13 104029

Larsluph:
We are currently reworking the whole project so the translation system will be rewriten as well. We will include a way to parameterize translation keys and prevent these issues in the future

Moius:
> Also I have concerns about the translation in line 138, as mentioned in #594. I believe it would be better to leave it unchanged to avoid this awkward appearance when selecting torrents. 螢幕擷取畫面 2023-04-13 104029

i have edit the code let it like this:
图片
should i commit to the main?

Larsluph:
> I achieved the above effect by modifying Getters.ts, but this should only be a temporary solution. Can I merge it?

You can always create a PR. Don't worry if it's only a temporary fix, as I just told we are rewriting everything so it will be improved in the v2 anyway

Moius:
> 

THANKS, i have created a PR for this.

PR #765, opened 2023-04-13
fix: fix the Chinese word order problem

Moius:
# fix: fix the Chinese word order problem

Modified part of the problem that the display is correct in English but the word order is wrong in Chinese. Both Simplified Chinese and Traditional Chinese are supported, but this is a temporary fix. Because this problem may not exist after refactoring.

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/lang/zh-hans.json
+++ b/src/lang/zh-hans.json
@@ -46,7 +46,7 @@
   "feed": "订阅源",
   "rule": "规则",
   "then": "时",
-  "of": "of",
+  "of": "中的",
   "yes": "是",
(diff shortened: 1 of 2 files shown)


Moius:
@WDaan Hello! If this PR is not merged, then #766 should not be merged either. This PR fixes the displaying inflection problem between Simplified Chinese and Traditional Chinese by modifying the getters.ts file, and only modifies the translation of Simplified Chinese to match the inflection, #766 modifies the translation of Traditional Chinese to match the inflection.

PR #770, opened 2023-04-16
Sync changes

LeaderbotX400:
Complete moving the UI of the context menu to Vue 3.
Start to scaffold logic for the context menu.


src/components/__tests__/HelloWorld.spec.ts
WDaan, line 1: Can we sync up so changes like these are no longer included 😛 It's a bit tedious to keep removing them

WDaan (changes requested):
Awesome work, once again!

Making great progess, it's going smoother than expected 🚀 

Once we move bigger parts like the TopActions with all the modals + the settings page, we will probably move some parts around in a way that makes the most sense... so don't be scared to move things around

LeaderbotX400 (commented):
Looks good to me

Issue #96, opened 2021-01-09
[Feature Request] Watch Folder


mivadebe:
First of all, I love the interface! Have been using it for a couple of weeks now. 
One thing I was missing is the settings for defining the monitored folders / watch folders. Would be great if that could be added.

WDaan:
Thanks for the kind words!

I've been trying to add it because it seemed like a very trivial thing to do ... but I can't get it to work. (by using their API)
I tried to change the monitored folder from the default WebUI and it doesn't seem to work either :slightly_frowning_face: 

Has that ever worked for you?

mivadebe:
Hi

It works for me, but only when i add multiple folders. Seems like it doesnt
get the first one somehow. Add the same folder twice and it will work.
Though, in the config file it shows a strange path..

Keep up the nice work!

Gr
Michel

WDaan:
![image](https://user-images.githubusercontent.com/25707052/104121152-8afe5900-533c-11eb-8fc4-2b651ba26b78.png)

Still doesn't seem to do anything :sweat_smile:

mivadebe:
Hi,

See below my settings & docker image I use.


[image: image.png]

I use the following docker image with OPENVPN included:

docker-compose:

# QBITTORRENT
  qbittorrent-openvpn:
    container_name: qbittorrent-openvpn
*    image: guillaumedsde/alpine-qbittorrent-openvpn:latest*
    restart: always
    cap_add:
      - NET_ADMIN
    volumes:
*      - "/media/downloads:/downloads"*
      - "${DOCKERDIR}/qbittorrent/config:/config"
      - "/etc/localtime:/etc/localtime:ro"
    environment:
      - PUID=${PUID}
      - PGID=${PGID}
      - TZ=${TZ}
      - OPENVPN_PROVIDER=###
      - OPENVPN_USERNAME=###
      - OPENVPN_PASSWORD=###
      - OPENVPN_CONFIG=###
      - LAN=192.168.1.0/24
    ports:
      - "8080:8080"
#      - "9117:9117" # jackett
#      - "7878:7878" # radarr
#      - "8989:8989" # sonarr


Hopefully that helps:)

WDaan:
Thanks! I got it to work on the default WebUI, however their API is super weird for this.
Haven't quite gotten it to work yet

PR #793, opened 2023-05-01
Update settings

### LINKED ISSUE #605: VueTorrent missing advanced settings
The plain Qbittorrent UI has a tab under settings called Advanced Settings which has, for example, an option to let you choose which network interface you want to use (important for VPN). That tab seems to be absent when using the VueTorrent UI. Am I missing something or is it not available?

Larsluph:
# Update settings [perf]

- [x] VueTorrent
- [x] Downloads
- [x] Connection
- [x] Speed
- [x] BitTorrent
- [x] RSS
- [x] WebUI
- [x] Advanced

# PR Checklist

- [x] All Unit tests pass

Fixes #605


WDaan:
Awesome!

After all this time we're finally on track to finish up all settings pages 😅

Larsluph:
@WDaan Everything should be good, just need to review some elements for better responsive but I'll do that in a second time 😄
The only thing that is missing is the famous monitored folders (cc #96)

WDaan:
Awesome! Thanks for putting in all the effort. Let's merge this version so we can test-drive it a little bit.
We can make those little tweaks as we go!

Monitored folders is a pain, I was hoping that by now they would've updated the API to be a little bit more userfriendly but no luck..


Thank you for your continued efforts in making this the best alternative webui on the market 😄

PR #800, opened 2023-05-04
feat: limit updateMainData to one request at once

tianze0926:
This pull request limits `updateMainData` requests to at most one at the same time.

On my Raspberry Pi it can happen that a request takes longer than the interval (2 seconds in [current implementation](https://github.com/WDaan/VueTorrent/blob/2fdeb63c83eee6bbcdcc0bc639a4d3d303f10e20/src/store/actions.ts#L12)) when qBittorrent is busy. The current polling method, which makes a request every 2 seconds without exception, can cause multiple identical requests, resulting in even longer response time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "vuetorrent",
-      "version": "1.5.5",
+      "version": "1.5.6",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


WDaan:
@Larsluph  any objections for this one?

Larsluph (approved):
LGTM

PR #811, opened 2023-05-12
perf: Extend RSS Feeds / Rules to use full width instead of half

Larsluph:
# Extend RSS Feeds / Rules to use full width instead of half [perf]

From this:
image

To that:
image

# PR Checklist

- [x] All Unit tests pass


WDaan:
Looks way better 👍

PR #817, opened 2023-05-15
[Vue 3] Global modals management

k4lu-0p:
Migrate global modals management to Vue 3


WDaan:
Awesome! Thanks for the contribution

PR #818, opened 2023-05-18
[Vue 3] Feat: add torrents modal

k4lu-0p:
Allows you to add one or more torrents to the queue via the AddModal.vue component

### DIFF
--- a/src/Pages/Dashboard.vue
+++ b/src/Pages/Dashboard.vue
@@ -15,21 +15,15 @@ definePage({
 const mainData = useMainData()
 
 const torrents = computed(() => {
-  const torrents: Torrent[] = []
+  // temp
+  const _torrents = Object.values((mainData.data.value?.torrents as any))
 
-  // const ts = mainData.data.value?.torrents
-  // for (let t in ts) {
-  //   let data = ts[t]
-  //   console.log(data)
-  //   // torrents.push(new Torrent({ ...data }))
-  // }
+  // if (isProduction()) return torrents
+  // if (import.meta.env.VITE_USE_FAKE_TORRENTS === false) return torrents
+  // const count = import.meta.env.VITE_FAKE_TORRENT_COUNT
+  // torrents.push(...generateMultiple(count))
 
-  if (isProduction()) return torrents
-  if (import.meta.env.VITE_USE_FAKE_TORRENTS === false) return torrents
-  const count = import.meta.env.VITE_FAKE_TORRENT_COUNT
-  torrents.push(...generateMultiple(count))
-
-  return torrents
+  return _torrents
 })
 </script>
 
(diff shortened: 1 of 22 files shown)


src/Pages/Dashboard.vue
```diff
@@ -15,21 +15,15 @@ definePage({
 const mainData = useMainData()
 
 const torrents = computed(() => {
-  const torrents: Torrent[] = []
+  // temp
+  const _torrents = Object.values((mainData.data.value?.torrents as any))
```
WDaan, line 19: Cool, but we should try to keep the typings as much as possible

src/composables/api/torrents.ts
```diff
@@ -1,19 +1,115 @@
...
+
+export const useTorrents = () => {
+
+  async function addTorrents(payload: AddTorrentsPayload): Promise<any> {
+    
+    let requestBody: FormData | URLSearchParams | null = null;
+    const files: File[] = payload.files
+
+    // Map params
+    const params: QbAddTorrentPayload = {
```
WDaan, line 59: This can probably be done a bit more efficient

WDaan (commented):
Sorry this took a while, public holidays and bussiness trip...

Awesome! thanks for picking this up. Can't wait to try this out.
We're still quite far away from actually moving over and releasing but at least we're making some progress.

src/composables/api/torrents.ts
```diff
@@ -1,19 +1,115 @@
...
+      autoTMM: payload.autoTMM,
+      category: payload.category?.name,
+      cookie: payload.category?.name,
```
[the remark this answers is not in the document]
WDaan: Thanks @Larsluph for taking a look at this as well

src/components/Modals/AddModal.vue
```diff
@@ -1,62 +1,354 @@
...
+const params = reactive<AddTorrentsPayload>({
+  urls: '',
+  files: [],
+  tags: [],
+  category: undefined,
+  directory: DEFAULT_DEST_FOLDER_PATH,
+  start: false,
+  autoTMM: false,
+  contentLayout: AppPreferences.ContentLayout.ORIGINAL,
+  stopCondition: AppPreferences.StopCondition.NONE,
```
Larsluph, line 51: Default value for these 3 is in settings

src/components/Modals/AddModal.vue
```diff
@@ -1,62 +1,354 @@
...
+
+const isFormValid = ref(false)
+const isDropAreaVisible = ref(false)
+
+const params = reactive<AddTorrentsPayload>({
+  urls: '',
+  files: [],
+  tags: [],
+  category: undefined,
+  directory: DEFAULT_DEST_FOLDER_PATH,
```
Larsluph, line 47: for this one, it would be awesome to update the value when the category is changed

Larsluph: mb, haven't seen the watch ^^

src/composables/api/torrents.ts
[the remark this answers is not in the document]
k4lu-0p: Oops! Thanks @Larsluph

WDaan:
Why closing? 😛 I'm sure we can use most of it?

k4lu-0p:
Sorry, issue with Git in my code editor, i'm fixing it

src/composables/api/torrents.ts
k4lu-0p: Qbit payload properties do not have a constant format. There's camelCase, snakeCase and so on... I can't see any other way of solving this problem. Do you have a suggestion?

src/components/Modals/AddModal.vue
k4lu-0p: settings/preferences are loaded when the modal has been opened

PR #819, opened 2023-05-18
perf: Rework search engine

Larsluph:
# Rework search engine [perf]

It's missing column filters to replicate the base WebUI if we want to be perfect but everything essential is there and working!

# PR Checklist

- [x] All Unit tests pass


WDaan (commented):
Awesome! Nice cleanup on the api side. What's holding this back?

Larsluph:
> Awesome! Nice cleanup on the api side. What's holding this back?

Got some exams right now, I just need to take time to design the view

WDaan:
No worries! Good luck 🤞

WDaan (approved):

WDaan:
Awesome work! I really like the new plugin manager, super sleek. 🚀 

Only remark I can find is, on PWA mobile... the search table doesnt fit the screen so you still need to horizontolly scroll to be able to hit the download button. Don't really understand why as putting it on mobile screen size using devtools does not cause this issue.

Either way, great work! Huge improvement over the current one.

Larsluph:
haha responsive design as always 😁, we have the problem on every other data table we're using, I think you can fix it using `width: 100%`. Will check that later to fix that on rss articles as well

PR #844, opened 2023-06-01
perf: Merge MatchingArticles into RuleForm

### LINKED ISSUE #833: [Feature request] RSS Downloader rules creator in real-time
It would be nice to be able to see in real-time matching entities for selected feeds when creating/editing the rule for RSS Downloader. It works the same way on desktop GUI. This feature is very convenient as it allows to check if your rule is correct (especially while using regex).

Larsluph:
# Merge MatchingArticles into RuleForm [perf]

- Merged MatchingArticles modal into RuleForm modal
- Added save FAB to prevent scrolling to the bottom and ease quick edit for preview. (FAB is out of the modal for now as it is a fixed display, don't know if it can be improved atm)

Note: Disabled matching articles on xs devices and portrait, no idea how we are going to make it work flawlessly on small screens

Fixes #833

# PR Checklist

- [x] All Unit tests pass


WDaan:
Allright, great start!

Yes, getting a complex UI working on the smaller form factor is always tricky. You either build whole new UX which causes a lot of overhead or you're always running into small issues/glitches.

Larsluph:
yep that's why I kept it on landscape orientation so that it isn't too much effort if you only have a small device 😁

PR #848, opened 2023-06-04
fix: context-menu not appearing

BlockListed:
# Fix context-menu not appearing [fix]
The body contains `oncontextmenu="return false"`, which, at least in all my test environments, causes all context menus to be blocked. This removes the ability to paste magnets on android (not sure about IOS).
The context-menu blocking is already handled by `blockContextMenu()` in `App.vue`, which handles TEXTAREAs etc. correctly, so this shouldn't break anything (according to my testing).

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,7 +14,7 @@
     <meta name="theme-color" content="#000" />
     <title>qBittorrent</title>
   </head>
-  <body oncontextmenu="return false">
+  <body>
     <noscript>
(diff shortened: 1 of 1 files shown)


Larsluph:
Awesome! It doesn't seem to break the actual block of context menu which is perfect 👌

Larsluph (approved):
Awesome! It doesn't seem to break the actual block of context menu which is perfect 👌

Issue #863, opened 2023-06-15
Allow using native find


nattyg93:
Hi folks,

It would be good if we could still use the native find functionality. On Mac in Firefox, if I hit "cmd + F" it goes to this app's filter bar, which is great, but sometimes I just want to use the native find functionality.

One option might be to open the native search if you hit "cmd + F" again while the app's search input has focus. This is what the [stripe api docs](https://stripe.com/docs/api) do for example:

<img width="532" alt="Screenshot 2023-06-15 at 12 45 17" src="https://github.com/WDaan/VueTorrent/assets/4318027/a0208032-bb6f-4966-8934-834af855dbf8">

PS. I love this project! I often use it on mobile when I'm out and about and its responsive design saves me such a hassle.

PR #864, opened 2023-06-15
perf: Allow native find

Larsluph:
# Allow native find [perf]

If ctrl+F / cmd+F is pressed and search input is already focused, will trigger native find

Note: Nothing is told inside the app, I don't know if it's necessary and where it should be said.

# PR Checklist

- [x] All Unit tests pass


WDaan:
#863?

We could add it to the readme :)

Larsluph:
yep forgot to mention it for auto close 😅

PR #940, opened 2023-07-05
perf: Rework Content tab

### LINKED ISSUE #304: Do not prefetch files in torrent's content
As the title, i know prefetch give it somewhat smoother experience, BUT i have a torrent with more than 3000 files and when i clicked "Show info" (to view peers, etc.) it crash the webui immediately.

The solution can be only load the "Content" tab when user click it and have it lazy load, only load more if the user scroll to the end of current view.

Other thing: Please add the search to the "Content" tab too (like the original UI).

### LINKED ISSUE #619: Subdirectories collapsing when changing file priorities
When I look at the content of a torrent, I may expand a directory to reveal season episodes, for example.
If I then change the priority of a single episode to "max" or whatever, the directory tree collapses back down again.

### LINKED ISSUE #696: Individual content percentage does not automatically update [1.5.0]
chrome_whKhwKLfO0

Shouldnt this be automatically updated? It does update when i press F5 to reload whole page but it does not update as im looking at it. It does update also if i switch between tabs in torrent detail

### LINKED ISSUE #703: Renaming content causes the "item wanted" check mark to be deselected.
When renaming a file/folder, after pressing the "save" button, then changing Vue tabs (eg. go to Info, then back to content), then the renamed file/folder is no longer checked as wanted.

Tested on Mobile and Desktop Chromium with no extensions enabled.

Larsluph:
# Rework Content tab [perf]

Complete rework of the tree structure to handle observers not working previously.

I had to change rename method to modal or the auto update could not work.

The only issue with the new system is the virtual DOM, when expanding/collapsing huge tree nodes the creation / deletion of a lot of elements in the DOM creates a huge lag spike. As stated in 

Fixes #304 (tested on a torrent with 3528 files and a maximum depth of 4)
Fixes #619
Fixes #696
Fixes #703

# PR Checklist

- [x] All Unit tests pass


WDaan:
Interesting! Would have to play around with it a little bit to figure this out.

WDaan:
I think that might be a good idea. We probably will get the question sooner or later.

You had this in a previous version of this PR, no? Why did you choose to step away from it?

WDaan (approved):

Larsluph:
> You had this in a previous version of this PR, no? Why did you choose to step away from it?

I'm not sure to follow you on that. For now to refresh I use the same interval as the global API I've done in a previous PR but that's it 🤔

WDaan:
Aha apologies! I probably got it confused with the API interval that's now configurable 😅

Yes, making this interval configurable would be neat. Implement if you want, otherwise we'll get to it if the need arises.

Larsluph:
It's quick and we've got nothing to lose so let's go! 🚀

WDaan:
Thanks for the quick turnaround :)

PR #956, opened 2023-07-11
chore: Revert sass bump

Larsluph:
# Revert sass bump [chore]

Revert SASS bump from dependabot (#885) to remove deprecation warnings.
Deprecation has been fixed with Vuetify 3. Update to Vue 3 will fix it as well.

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "eslint-plugin-vue": "^9",
         "jsdom": "^22.1.0",
         "prettier": "^3",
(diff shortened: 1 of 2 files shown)


WDaan:
Thanks! Has been annoying me as well.

There's no real upgrade path unless moving to 
vue3. Which is basically a complete rewrite. This will take a lot of time and effort for unfortunately very little difference in the end :/

PR #1008, opened 2023-07-27
fix: App not redirecting to login page on load

Larsluph:
# App not redirecting to login page [fix]

When old session is expired, state is still authenticated so no redirect was made when the app was reloaded.

# PR Checklist

- [x] All Unit tests pass


WDaan:
Good catch, probably left over from me fixing the signout bug

PR #1083, opened 2023-09-04
fixing formatDataValues for values with 4 digits or more fixes #1079

### LINKED ISSUE #1079: "Free Space" formated strangely
Really trivial issue.  My free space in the sidebar currently reads "1.00e+3" GB (it's actually 999.9 Gb) so I guess that should be 999.9GB or 1.00TB

jagadam97:
# fixing formatDaataValues for values with 4 digits or more fixes [fix]

in formatDataValues function we are using toPrecision funtion with precision value as 3 which when round off value reaches 4 digits returns exponential values. Fixed this by using checking if he value is above 999 and giving precision value 4 if the same is true.

Gave value :  
Screenshot 2023-09-04 at 06 28 37

function returned : 
Screenshot 2023-09-04 at 06 28 43

# PR Checklist

- [x] All Unit tests pass

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -24,7 +24,9 @@ export function formatDataValue(data: number, isBinary: boolean) {
   while (data >= base ** i) {
     i++
   }
-  return toPrecision(data / base ** (i - 1), i > 1 ? 3 : 1)
+  let value = data / base ** (i - 1)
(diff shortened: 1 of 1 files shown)


Larsluph (approved):
Good job finding that!👍

PR #1099, opened 2023-09-14
Revert "fix(logs): Wrong date parsing"

Larsluph:
Reverts WDaan/VueTorrent#1076


Larsluph:
I don't know if it's possible to add a note to the changelog about the qbit version change...

PR #1151, opened 2023-10-15
vue3-infinite-scroll

jagadam97:
# Infinite scroll [feat]

This PR adds infinite scroll feature for vue3 branch.
Not able to make the page size dependent on store value as on refresh the value is reseting to 0

# PR Checklist

- [*] I've started from Vue3
- [*] I've only committed changes related to this PR
- [*] All Unit tests pass
- [*] I've removed all commented code
- [*] I've removed all unneeded console.log statements

### DIFF
--- a/src/components/Settings/VueTorrent/General.vue
+++ b/src/components/Settings/VueTorrent/General.vue
@@ -6,7 +6,7 @@ import { computed, onBeforeMount, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { toast } from 'vue3-toastify'
 
-const paginationSizes = ref([5, 15, 30, 50])
+const paginationSizes = ref([0, 5, 15, 30, 50])
 const settingsField = ref('')
 
 const isProduction = computed(() => process.env.NODE_ENV === 'production')
--- a/src/pages/Dashboard.vue
+++ b/src/pages/Dashboard.vue
@@ -99,7 +99,7 @@ const {
   lastPage
 } = useArrayPagination(filteredTorrents, {
   currentPage: _page,
-  pageSize: vuetorrentStore.paginationSize
+  pageSize: vuetorrentStore.paginationSize === 0 ? 999999 : vuetorrentStore.paginationSize //TODO: Change the pagination size to No. of torrents filtered
 })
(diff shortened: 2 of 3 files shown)


WDaan:
Hi! Good idea.
Just a thought, couldn't you disable page size altogether if you want infinite scroll enabled?

jagadam97:
https://drive.google.com/file/d/1y2hvwHk2D4xUBxQdjiuSlPK5tgJUSKTl/view?usp=drive_link

jagadam97:
@WDaan sure will revisit

jagadam97:
@WDaan 
-> If i dont send the pageSize (key: value)its defaulting to 10 - 10 torrents shown but page number changes to 1 at every refresh
-> If i send pageSize as undefined its defaulting to 0 - no torrents shown
-> if i use filteredTorrents.value.length as pagesize after refresh this value is resetting to 0
till now what ever logic i tries the above logic only worked properly till now. assuming no one crazy enough to maintain 7 figures of active torrent on there server

Any modification of above commit is welcomed to make it better

jagadam97:
let me try few more things before i give up on this

Larsluph:
I think this isn't the correct way to do it, maybe try differenciate the paginated components and the virtual scroll component (using v-virtual-scroll from vuetify). The problem you're facing may come from the check I have put in place to prevent the page being 0 which is a bug of the pagination composable we use.

https://vuetifyjs.com/en/components/virtual-scroller/

https://github.com/WDaan/VueTorrent/blob/48dc583e354625e4e102b06839fbb1fd387a512c/src/pages/Dashboard.vue#L92-L95

src/components/Settings/VueTorrent/General.vue
```diff
@@ -6,7 +6,7 @@ import { computed, onBeforeMount, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { toast } from 'vue3-toastify'
 
-const paginationSizes = ref([5, 15, 30, 50])
+const paginationSizes = ref([0, 5, 15, 30, 50])
```
Larsluph, line 9: And for the setting, try something like that (create a locale key for that obviously)

```suggestion
const paginationSizes = ref([{ title: 'Infinite scroll!!', value: 0 }, 5, 15, 30, 50])
```

PR #1166, opened 2023-10-17
fix(pagination): Fix pageCount not being reactive

### LINKED ISSUE #1157: [Vue3 Dashboard] Pagination not reactive anymore
The pagination component's length only updates after unmounting / remounting Dashboard component.
Everything else works fine.

Larsluph:
Fixes #1157


WDaan:
Good catch

PR #1177, opened 2023-10-19
perf(keybinds): Improve keyboard events

Larsluph:
- Fix cmd key detection on mac devices
- Handle missing focus events on dashboard search input


WDaan:
@Larsluph any changes to the README required?

WDaan (approved):

Larsluph:
Good point, will add that 👀

PR #1178, opened 2023-10-19
feat(font): Self-host roboto font

Larsluph:


package.json
```diff
@@ -14,6 +14,7 @@
     "coverage": "vitest run --coverage"
   },
   "dependencies": {
+    "@fontsource/roboto": "^5.0.8",
     "@fontsource-variable/roboto-mono": "^5.0.14",
```
WDaan, line 18: should we remove this one?

WDaan (changes requested):

package.json
Larsluph: These two are different fonts, the latter is used for the cards in the navbar, I think we can remove them

WDaan: Yeah they provide cleaner numbers :)
`@fontsource/roboto` will probably still work

PR #1179, opened 2023-10-19
perf(build): Import stores from source instead of index.ts

Larsluph:
This removes all warning from vite which may cause circular dependencies while chunking.


WDaan:
Sad, single-line imports for all of them look cleaner

Larsluph:
Yeah haven't found a way to chunk it manually

PR #1181, opened 2023-10-19
feat(dashboard): Add compact mode

### LINKED ISSUE #157: [Feature Request] Minimal/Overall view
Hi,

Sometimes I have a lot of torrent at the same time. I'd like to have the possibility to have an overall view of all my torrents on the main page maybe with less metrics displayed.

Cheers !

Larsluph:
First try at the compact mode, it only displays torrents' title with its progress bar.
Waiting for feedback to improve 👀

Fixes #157

### DIFF
--- a/src/components/Dashboard/DashboardItems/ItemPercent.vue
+++ b/src/components/Dashboard/DashboardItems/ItemPercent.vue
@@ -4,7 +4,7 @@ import { formatPercent } from '@/helpers'
 import { Torrent } from '@/types/vuetorrent'
 import { computed } from 'vue'
 
-const props = defineProps<{ torrent: Torrent; title: string; value: string }>()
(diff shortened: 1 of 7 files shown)


WDaan:
Will take a look later today!

I was thinking about using https://tanstack.com/table/v8 for this, it would allow us to create a slim table, with great filtering and sorting capabilities. It's also a bit thé standard to use in complex tables. 

Maybe we could make this work for the default view too, solving those filtering/sorting capabilities people have been asking for

Larsluph:
The only problem with tables is the responsiveness, not sure that it'll be good looking on small devices :/

WDaan:
Yeah for mobile devices I'd still like to use cards, but tanstack does allow for a custom row renderer... so we should be able to get that card in there.

PR #1184, opened 2023-10-20
feat(dashboard): Add compact mode

Larsluph:
First try at the compact mode, it only displays torrents' title with its progress bar.
Waiting for feedback to improve 👀

Rebased #1181 after vue3 merge


WDaan (approved):

tainewoo:
if the columns can be selected by uers, I will try the table mode.
Table mode would be more effcient for users have thounds of torrents seeding/running.

Larsluph:
It's planned, I just need to rework some things before 😁

Larsluph:
Everything worked on my devices, might need to do a quick check on iOS PWA but everything should work as expected.

WDaan:
This is very cool! 🚀  Awesome 😄 

I do wonder about the usefulness of the grid view? It's like the list view, but more complicated to look at 😅

Larsluph:
It allows a more compact view than a full row if you only want 2 / 3 properties

Larsluph:
But yeah it's not great with lots of properties

FluffyDiscord:
The table view rows could be slimmer, is it too late?

Larsluph:
Table row height are determined based on the max height of each cells. If you want an improvement, you can still open an issue 😉

PR #1187, opened 2023-10-21
chore(faker): Improve torrent mocking

Larsluph:
- [X] Move faker code to torrent builder
- [x] Add static torrent mock handling


WDaan:
Nice, this is better. Ideally the behave more real as well... metadata stays the same and only the progress and speeds jumps around a bit.

WDaan:
@Larsluph can we merge this one? Don't think it's blocking anything

Larsluph:
I just want to test one last thing before and we can merge it

PR #1202, opened 2023-10-24
feat(filters): Allow multiple selection

### LINKED ISSUE #1165: [Feature Request] Adding new status filter: Uploading
Since we have a new status of "Uploading" means active uploading status for torrents in the torrent list view, can we also add this to the status filter? That would be great.

Currently we have the "Active" and "Downloading" filter.
The "Active" filter lists all the active uploading and downloading torrents.

Larsluph:
⚠️ Breaks navbar display with the rework of dashboard store structure

Fixes #1165
Fixes Larsluph#8


src/styles/colors.scss
```diff
@@ -8,21 +8,6 @@ $category: #04669a;
 $tag: #048b9a;
 $tracker: #c97d09;
 
-// State filter colors
```
WDaan, line 11: why get rid of the colors?

WDaan (approved):
Neat! Only a bit unfortunate that you have to clear your local storage to get it to work. Which for mobile devices probably means a reinstall of the PWA 🙂

Larsluph:
Yeah will check later to prevent the issue because it causes the top bar to not display so you can't go to settings using the UI, you will need to go to the settings route.

src/styles/colors.scss
Larsluph: It was only used for the active filters state chip, which isn't needed anymore because we now filter by torrent state and values differs between these 2

WDaan:
Yeah i cleared it using the console, but that's not something you can do on a phone (and isnt a very good experience in general)

Thanks for taking a look!

Larsluph:
@WDaan Only way I found to fix that is to change the key 😄. Also took the opportunity to put it in a proper store.

WDaan:
Great! Will take a look tomorrow

WDaan:
This looks amazing!! 🚀 very cool

Screenshot 2023-10-26 at 11 53 01

I think this chip should have a 'clear all filters'-option
And perhaps also a clear button per filter type? 🤔

Larsluph:
That can be a great final touch for that feature indeed! Prepend icons and listeners are already available on chips so it can be done really quickly 🚀

Larsluph:
Chips even have a closable attribute that shows a remove icon, callbacks are missing for now, will add them later.

PR #1213, opened 2023-10-28
fix(AddTorrentDialog): Fix persistence not working

Larsluph:


Larsluph:
File persistence still doesn't work but I think that comes from the File implementation

Issue #1247, opened 2023-11-05
[Bug] General Settings “Current Version” display incorrectly 


bolucat:
One hour ago, I switched to the latest release [v2.0.1](https://github.com/WDaan/VueTorrent/releases/tag/v2.0.1), but after I cleared the browser cache, the “Current Version” in general settings still shows “2.0.0”. I opened a private window to make sure I don't use the stale browser cache, but nothing changes. 

So I searched the string “2.0.1” around all the files in “public” directory, but there's nothing. So I searched the string “2.0.0” and found it exits in `public/assets/Settings-8f4a6f00.js`
```bash
......
g=Vt(()=>t.value?"2.0.0":s.value?"DEV":null)
......
```
So maybe there's a bug that the version value in `package.json` didn't pass correctly to the final JS files

Larsluph:
That seems to be a CI issue. By checking the logs it's releasing the commit just before the version bump.

PR #1248, opened 2023-11-05
chore(ci): Fix version not up-to-date in release build

Larsluph:
Not sure if it'll do the trick but the problem is that the checked out ref is the one before the release.

It can be an async problem between jobs (which the `needs` should fix) or a default value problem (fixed in this PR)

#1247


WDaan:
Let's find out :)

PR #1259, opened 2023-11-05
feat(title): Ability to set custom browser tab title

### LINKED ISSUE #1005: [Feature Request] Customizable title either as HTML header or as header bar
Hi VueTorrent team. Thank you so much for this amazing WebUI. I switched all my instances' webui to this the first day I found this repo 😊 Kudos!

I had two qBittorrent instances that I was actively using, and as I got into cross-seeding recently, that number jumped to 4. It has gotten hard for me to distinguish between instances: I can't figure out which is which quickly since the tab names are the same. Even glancing at the home page doesn't help and I'll have to resort to my long-term memory and the IP address.

I am proposing a feature that lets us customize either the HTML title(browser title bar) or the "qBit**torrent**" text in the title section of the UI.

Thanks! Kudos again for such a great WebUI!

Larsluph:
Fixes #1005

@luigidotmoe

### DIFF
--- a/src/components/Settings/VueTorrent/General.vue
+++ b/src/components/Settings/VueTorrent/General.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { titleOptionsList } from '@/constants/vuetorrent'
+import { TitleOptions } from '@/constants/vuetorrent'
 import { LOCALES } from '@/locales/locales'
 import { useAppStore } from '@/stores/app'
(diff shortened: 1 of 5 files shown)


N0rdmann:
Could the "custom title" field also be enabled for the 2 speed options instead of the fixed: [D: , U: ] VueTorrent? 😇
Keep up the great work, Cheers 👌

Larsluph:
> Could the "custom title" field also be enabled for the 2 speed options

Not sure what would be the point of that...

Issue #1266, opened 2023-11-06
Search engine shortcut regression


Anteus:
Shortcut to open search engine by pressing "/" like before:
https://github.com/WDaan/VueTorrent/blob/3c9f4e3ad59a88a9b8ce45dec40b72877c285513/src/views/Dashboard.vue#L490

Could be added here:
https://github.com/WDaan/VueTorrent/blob/908ef994711c2f588099880841421eb8b7eafcd5/src/pages/Dashboard.vue#L178

PR #1269, opened 2023-11-06
#1266 search shortcut

Anteus:
#1266 search shortcut regression fix

Please explain what kind of feature your PR's brings or what kind of bug it fixes. Feel free to include screenshots for UI related changes. Keep all the commits inside your PR
related to topic of the PR, otherwise create a seperate PR. Don't forget to link the related issue! (if applicable)

## PR Checklist

- [ ] All tests pass

### DIFF
--- a/src/pages/Dashboard.vue
+++ b/src/pages/Dashboard.vue
@@ -221,6 +221,16 @@ function handleKeyboardShortcuts(e: KeyboardEvent) {
     e.preventDefault()
     return true
   }
+
+  // 'Search' => Search view
+  if (e.key === '/') {
+    const searchInput = document.getElementById('searchInput')
+    if (document.activeElement !== searchInput && !isSearchFilterVisible.value) {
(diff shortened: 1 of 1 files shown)


src/pages/Dashboard.vue
```diff
@@ -221,6 +221,16 @@ function handleKeyboardShortcuts(e: KeyboardEvent) {
     e.preventDefault()
     return true
   }
+
+  // 'Search' => Search view
+  if (e.key === '/') {
```
Larsluph, line 226: lgtm, was just wondering if we should include the ctrl/command key to that keybind

Anteus: Have to use shift+7 already on my current keyboard to get "/" so would be a bit wonky with ctrl+shift+7. Just / is the same shortcut in Gmail. Github supposedly also have "/" but that one doesn't work on my current layout only the alternate keyboard shortcut "s" works.

PR #1276, opened 2023-11-06
perf(localization): Integrate Tolgee inside CI

Larsluph:


WDaan:
Awesome! Great step in the right direction, next improvement would be to pick up new available languages automatically.

luigidotmoe:
Good job, great new feature

PR #1286, opened 2023-11-09
fix(settings): Update drag handling to completed properties

### LINKED ISSUE #1243: [Bug] Torrent Card on Android has become impossible to scroll
After updating to 2.0.0 I lost my order preferences so I am now trying to adjust the Torrent Card on mobile. Specifically my Android 13 device running Firefox 119.0.1.

The problem I'm running into is the drag feature to reorder these fields is interfering with scrolling and taking precedence over it. There is nowhere I can tap that will allow me to scroll. Instead it just selects the field and starts dragging it around for reordering. See screenshot below.

Maybe we can find a way to where the drag feature only activates when tapping on or near the two columns of little dots? Or allow scrolling in the blank space to the right?

Btw, thank you so much to Daan and the team for your dedication to this project. It's immensely useful to me and I use it every day.

Screenshot_20231104_170310_Firefox

Larsluph:
Fixes #1243


WDaan (approved):
You know you're allowed to sleep, right?

PR #1294, opened 2023-11-10
fix(Navbar): Add missing bind on right drawer setting

### LINKED ISSUE #1291: `Right Drawer` has no effect
image

The `Right Drawer` checkbox does not have an effect, the side bar remains on the left side. Can fix? Thank you.

Larsluph:
Fixes #1291


bilogic:
May I ask when this will be released? Thank you.

PR #1295, opened 2023-11-10
fix(RightClickMenu): Fix target not being selected on long press

### LINKED ISSUE #1290: Mobile View -Holding down on the options of one torrent affects all torrents on the client.
Hello

When I hold down on a torrent to pause or recheck it starts the selected process on all torrents. I ran into this issue by accident when I tried to recheck one specific torrent which started the rechecking process on all 2000+ torrents on my client.  
Vue2 1 0

I installed 1.8.0 on another client and tested it and it only makes the change on the selected torrent. 
Vue1 8 0

Not sure if this is by design due to the major update to 2.*, but are you guys able to change the functionality back to how it was in vue 1.8.0? 

Thank You

Larsluph:
This should fix the problem, can't test because right click works perfectly fine everywhere except on iPhones.

Fixes #1290

### DIFF
--- a/src/pages/Dashboard.vue
+++ b/src/pages/Dashboard.vue
@@ -326,7 +326,7 @@ function endPress() {
 
         <v-list-item v-for="torrent in paginatedTorrents" :id="`torrent-${torrent.hash}`"
                      :class="display.mobile ? 'mb-2' : 'mb-4'" class="pa-0" @contextmenu="onRightClick($event, torrent)"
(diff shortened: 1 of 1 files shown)


WDaan:
Will test :)

Larsluph:
@WDaan 
To test it you can just see if the torrent card highlights.

WDaan:
Seems to be working as expected on my iPad

PR #1299, opened 2023-11-12
chore(ci): Use custom action to pull changes from Tolgee on nightly build

Larsluph:
@WDaan Before merging, you'll have to [enable write-access on autogenerated github token](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository)

Tested on my fork and it works great! [link](https://github.com/Larsluph/VueTorrent/actions/runs/6842355392/job/18603741445)


WDaan:
I had a version with just an auto-commit but that didn't work because of the branch protection. Using a PAT would cause infinite CI.

Would have to take a bit of a closer look at this one.

Larsluph:
Arf yeah branch protection might break things. Will see if it can be bypassed in some way.

WDaan:
Yeah with PAT, but not ideal.

tolgee-action/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+export async function exportLocaleData(api_key: string) {
+  const curlPath = await which('curl', true)
+  const unzipPath = await which('unzip', true)
+
+  await exec(curlPath, ['-H', `X-API-KEY: ${ api_key }`, export_url, '-o', 'locales.zip'])
+  await exec(unzipPath, ['-o', '-d', './src/locales', 'locales.zip'])
+  await rm('locales.zip')
+}
+
+export async function commitChanges(committer_name: string, committer_email: string, commit_message: string) {
```
WDaan, line 68: You can probably replace this with just: https://github.com/stefanzweifel/git-auto-commit-action
in your action.yaml

tolgee-action/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+import { mkdirP, which } from '@actions/io'
+import { rm } from '@actions/io/lib/io-util'
+import * as fs from 'fs'
+import { LanguageResponse } from './types'
+
+const base_url = 'https://app.tolgee.io/v2'
+const languages_url = `${ base_url }/projects/languages`
+const export_url = `${ base_url }/projects/export?format=JSON&filterState=TRANSLATED%2CREVIEWED%2CDISABLED`
+
+async function extractProjectLanguages(httpClient: HttpClient, api_key: string) {
```
WDaan, line 13: Can't we just to the curl step like we did before? 🤔

tolgee-action/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  const projectLocales = r_lang.result._embedded.languages
+  return projectLocales.map(locale_metadata => ({
+    importSymbol: locale_metadata.tag.toLowerCase().replace('-', '_'),
+    enumName: locale_metadata.tag.toUpperCase().replace('-', '_'),
+    tag: locale_metadata.tag,
+    originalName: locale_metadata.originalName
+  })).sort((a, b) => a.tag.localeCompare(b.tag))
+}
+
+export async function updateLanguagesMetadata(httpClient: HttpClient, api_key: string) {
```
WDaan, line 30: This is really cool :)

tolgee-action/action.yml
```diff
@@ -0,0 +1,21 @@
...
+    default: 'github-actions[bot]'
+  author_email:
+    description: 'Email used when updating locale data'
+    required: false
+    default: 'github-actions[bot]@users.noreply.github.com'
+  commit_message:
+    description: 'Commit message used when updating locale data'
+    required: false
+    default: ''
+runs:
```
WDaan, line 19: Maybe we can use bun for this instead of node? That we we can skip all the transpile stuff and just natively run the typescript script.

WDaan:
This still won't work with the branch protection? 🤔
We can:
- make it PR
- make it commit to the release branch
- ...

tolgee-action/action.yml
Larsluph: I'm not aware of that possible, node20 is what is used in the gha environment.

Larsluph: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsusing-for-javascript-actions

tolgee-action/src/utils.ts
Larsluph: It just prevents having to save a file and stuff, we directly extract json from response and process it to get languages metadata

tolgee-action/action.yml
WDaan: Yes they try to 'help' you. But we can probably just do a custom action and run using bun
https://bun.sh/guides/runtime/cicd

We can definetely do it using a composite action as well.

tolgee-action/src/utils.ts
Larsluph: This function only downloads the languages metadata, real translations are downloaded later on in `exportLocaleData()`

Larsluph:
We can create a PR but everything else will be restricted by protection.

The only "clean" way I see would be to find a way to not trigger the CI on every push by including some exceptions (with a particular commit message for example) and use a PAT to bypass protection.

Issue #1285, opened 2023-11-09
[Bug] Getting login screen even when logged in


SinTan1729:
Even if I'm logged in, I get a login screen while accessing the VueTorrent page. Going to the same address again fixes it. I don't get asked for logins across reloads.

[Here's a screen recording.](https://streamable.com/ympfhc)

Larsluph:
This can happen sometimes. We can't replicate this bug so it's kinda hard to fix. If you have any idea, we'll take them!

WDaan:
I run into this as well. Even more so on PWA, we haven't been able to come up with a fix yet though 🙁 

Thanks for reporting!

zbikk:
can replicate, even with 0.0.0.0/0 bypassed i still get the login screen everytime, doesn't happen on native web ui.

Larsluph:
@zbikk If you have reliable reproduction steps (preferably from clean setup), we'll be happy to hear from you. It will greatly decrease the time required to fix it!

zbikk:
https://i.imgur.com/oanmDiL.mp4

as you can see from the video above when at the login screen if i press on anything really and then return it gives me the torrent screen so it's not an authorization issue and rather an issue with the login screen itself. as for reproduction steps i can't give you much except that it happens on clean installs as well, here are my web ui settings: 
![image](https://github.com/WDaan/VueTorrent/assets/145393776/a4a153ff-d6f7-4a78-8043-b77b7154d924)

Larsluph:
@zbikk Do you know if it also happens in a clean browser environment (in private mode for example)?

SinTan1729:
I just tested. It does happen in private mode for Firefox. But it doesn't seem to happen at all in Chromium (both incognito and otherwise). But it happens in Chromium based browsers on Android.

plexecutor:
I setup VueTorrent yesterday and noticed this issue, but only when connecting via reverse proxy. I use Nginx Proxy Manager and when I access via my public domain name for the first time during a new browsing session using Chromium based browsers (Edge, Chrome), instead of seeing my basic auth prompt that NPM is configured for, I get the VueTorrent/qBittorrent login prompt--which will not accept my credentials. Reloading the page prompts for NPM basic auth and then bypasses VueTorrent/qBittorrent auth using the whitelisted IP. Strangely, this is not a problem at all if I use incognito mode. Using the native qBittorrent webui sans VueTorrent, it behaves as expected.

Beet4:
[...]

PR #1316, opened 2023-11-16
Add config option to disable auth

nebulade:
VITE_DISABLE_AUTH=true can be set during build time to make all routes public

This is useful in a trusted or proxied environment.

The change came up for the Cloudron package https://git.cloudron.io/cloudron/qbittorrent-app


WDaan:
Shouldn't this be more usefull as a user setting? Now people will have to manually rebuild to be able to use it.

nebulade:
Haven't thought of that this way, probably because I looked from a packaging perspective. Is there a way to specify such client settings via a static config file on the server? I am not too familar with the vue-pwa part.

Larsluph:
I'm not sure of the point of that PR, if you configured your qBittorrent instance to bypass authentication by using your trusted method, you shouldn't be seeing that login screen anyway. If that's the case you may face the same issue as in #1285

PR #1319, opened 2023-11-17
Fix authentication race condition (#1285)

### LINKED ISSUE #1285: [Bug] Getting login screen even when logged in

KoHcoJlb:
Fixes #1285

### DIFF
--- a/src/pages/Login.vue
+++ b/src/pages/Login.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import PasswordField from '@/components/Core/PasswordField.vue'
 import { useAuthStore } from '@/stores/auth'
-import { onMounted, reactive, ref } from 'vue'
+import { onMounted, reactive, ref, watchEffect } from 'vue'
 import { useRouter, useRoute } from 'vue-router'
 import { toast } from 'vue3-toastify'
 import { useI18n } from 'vue-i18n'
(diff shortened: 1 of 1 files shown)


WDaan:
Could you explain the reasoning why this would solve the issue?

KoHcoJlb:
Reasoning would be that this issue happens because of race condition when we check authentication
i.e. we start checking authentication asynchronously [here](https://github.com/WDaan/VueTorrent/blob/e1a9622da929855c0123afac5f429f9cbf3761d6/src/App.vue#L52), then while request is in flight we pass both authentication checks [here](https://github.com/WDaan/VueTorrent/blob/575b071f6db9edc01412fab45109d3f4d203552d/src/plugins/router.ts#L15) and [here](https://github.com/WDaan/VueTorrent/blob/master/src/pages/Login.vue#L56) while isAuthenticated is still false
This change allows Login component to catch when authentication check completes and drive login flow forward

KoHcoJlb:
I've also tested this and can confirm that it solves the issue.

Larsluph (approved):
Awesome, thanks for the investigation and fix!

WDaan:
Makes sense! Thanks for the investigation :)

WDaan (approved):

KoHcoJlb:
Thank you back for this beautiful ui)

PR #1349, opened 2023-11-21
perf(localization): Add missing keys to Tolgee

### LINKED ISSUE #1347: After setting the language to Chinese, the left status bar did not translate properly into Chinese
After setting the language to Chinese, the left status bar did not translate properly into Chinese

Larsluph:
Fixes #1347


WDaan:
Thanks! :)

PR #1355, opened 2023-11-23
feat(Settings): Handle tab routes in settings

Larsluph:


WDaan:
Why is this better as a router? Components becoming too big?

Larsluph:
So that tabs can be loaded from url directly (after reloading for example)

Larsluph:
Nothing has really changed in the router, I just added two optional parameters so that I can store and load them to switch tab on mount.

PR #1364, opened 2023-11-28
chore(deps): bump 7 packages

Larsluph:
#1362


Larsluph:
Last package to update is Vite 5, breaking `vite-plugin-vuetify`

PR #1367, opened 2023-11-29
perf: add translations cron

WDaan:
Will create a PR like here: https://github.com/WDaan/VueTorrent/pull/1368

Simpeler solution than https://github.com/WDaan/VueTorrent/pull/1310

Fixes https://github.com/WDaan/VueTorrent/issues/1289


Larsluph:
The downloaded zip file is added to the PR, we might need to remove it before creating PR?

PR #1399, opened 2023-12-08
chore(ci): Clean reduce and update

Larsluph:
- Clean jobs
- Reduce CI usage by building project once for release and nightly
- Update actions


.release-please-manifest.json
```diff
@@ -0,0 +1 @@
+{".": "2.3.0"}
```
WDaan, line 1: Will it both update the package.json and this manifest file? 🤔

I think it would only do one.

WDaan (changes requested):
Looks great, but I would not change the release-please config 🤔

.release-please-manifest.json
Larsluph: Yeah the manifest is updated with the version bump, I don't really see why it is needed but without it the action step fails

Larsluph:
> Looks great, but I would not change the release-please config 🤔

I changed it to update the action config. Parameters passed as action variables aren't specified anywhere in the docs, I assume it has been deprecated since then. I can rollback that change.

WDaan:
All good 😄 let's see if we run into any issues

PR #1458, opened 2024-01-13
perf: Update style to remove weird spacing

Larsluph:


WDaan:
Screenshot 2024-01-14 at 12 07 07


Don't you think this is a little big? 🤔

Larsluph:
I prefer it big and without space than small and offset with empty space 😂
But we can adjust that if you want 👀

WDaan:
Idk. I'm not sure. This way we lose 2 rows of space? 

Also I think we can make the sort select a bit bigger again🤔

Larsluph:
> Idk. I'm not sure. This way we lose 2 rows of space? 

Only on desktop where there's already plenty of space

> Also I think we can make the sort select a bit bigger again🤔

I already adjusted the max width to not wrap on mobile, we could switch it to just width instead

Larsluph:
@WDaan Is there anything blocking?

PR #1468, opened 2024-01-18
perf(PWA): Use Network-only strategy for service worker

### LINKED ISSUE #1350: Strange behavior of login with authentication bypass
I use the hotio docker of qbt with the vue theme. It is behind traefik, protected by the Authelia SSO portal.
In settings/webui I selected "Bypass authentication for clients whitelisted IP subnets" and put in all private networks:
- 10.0.0.0/8
- 172.16.0.0/12
- 192.168.0.0/16

The goal was that qbt would not ask for a password at all, since the traffic always comes from a private network via traefik. And the UI is protected by Authelia.

This led to a very inconsistent experience.

- The first login on a fresh browser (nothing in cache/ no cookies) works always fine
- Every second visit of the page has mixed outcomes:
  -  Almost always I can see the sidebar (with actual stats). Sometimes with torrents in the main window, sometimes with the login mask. The menu on top is also visible. I can click on "log" and close it again and I see my torrent list. Putting in my credentials to the login mask usually leads to an error. Refreshing the page sometimes also removes the login and shows the torrents.

Larsluph:
Fixes #1350

### DIFF
--- a/public/registerSW.js
+++ b/public/registerSW.js
@@ -1,0 +1,1 @@
+"serviceWorker"in navigator&&window.addEventListener("load",(async()=>{await navigator.serviceWorker.register("/sw.js")}));
--- a/index.html
+++ b/index.html
@@ -13,6 +13,10 @@
     <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
     <meta name="theme-color" content="#000" />
     <title>VueTorrent</title>
+
+    <!-- PWA setup -->
+    <link rel="manifest" href="/manifest.webmanifest" crossorigin="use-credentials">
+    <script src="/registerSW.js"></script>
   </head>
   <body>
     <noscript>
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "pinia": "^2.1.6",
         "pinia-plugin-persist": "^1.0.0",
         "uuid": "^9.0.1",
-        "vite-plugin-pwa": "^0.17.4",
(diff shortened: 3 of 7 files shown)


WDaan:
🤔

WDaan:
What's the impact of this?

Larsluph:
There shouldn't be any impact, I built the project and copied the generated manifest and service worker from the Vite PWA plugin. And modified the service worker to disable caching.

After the investigation in the linked issue, caching the UI is causing some problems when using a reverse proxy and custom SSO auth.

WDaan (approved):

WDaan:
I do seem to notice a couple changes since this change.

But would need to deep dive if this is really the cause and how this change could have impacted it.

public/registerSW.js
ICE-GB, line 1: Should we use `. /`? I'm using the nginx proxy with path `a.b.c/qb/`

public/registerSW.js
```diff
@@ -0,0 +1 @@
+"serviceWorker"in navigator&&window.addEventListener("load",(async()=>{await navigator.serviceWorker.register("/sw.js")}));
```
ICE-GB, line 1: Should we use `. /` ? I'm using the nginx proxy with path `a.b.c/qb/`

Larsluph: I just created a PR to address that issue:

https://github.com/VueTorrent/VueTorrent/pull/1511

PR #1470, opened 2024-01-18
feat(content): Rework content tab entirely

### LINKED ISSUE #1456: Bug/Feature/Improvement: Content Tab
# Description

There are a lot of ergonomic improvements that could be made to the content panel, however we're waiting for [this](https://github.com/vuetifyjs/vuetify/issues/13518) component to become available

Grouping all Content related issues:
- #1432 
- #1363
- #1174
- #1173
- #1115
- #894 
- #1411
- #1019

Larsluph:
Fixes #1456


Larsluph:
@cybersholt Here's the current progress. Responsive design is still missing but at least desktop version looks nice.

image

Larsluph:
This misses sort and filter capability but everything should work fine. I sorted by name in the content store > flatTree computed

src/pages/Dashboard.vue
```diff
@@ -171,6 +184,262 @@ function handleKeyboardShortcuts(e: KeyboardEvent) {
...
 }
 
+const isMultiple = computed(() => dashboardStore.selectedTorrents.length > 1)
+const hashes = computed(() => dashboardStore.selectedTorrents)
+const hash = computed(() => hashes.value[0])
+const torrent = computed(() => torrentStore.getTorrentByHash(hash.value))
+const torrents = computed(() => dashboardStore.selectedTorrents.map(torrentStore.getTorrentByHash).filter(torrent => !!torrent))
+const availableCategories = computed(() => [{ name: '' }, ...maindataStore.categories])
+
+async function resumeTorrents() {
```
WDaan, line 194: Why not put all of this in the separate TRC component? 🤔 

Feels like it's poluting the Dashboard component quite a bit

WDaan (changes requested):
Content tab looks good! Also way more useable on mobile

src/pages/Dashboard.vue
Larsluph: I put it here because I made the TRC component generic to also use it in the content tab without duplicating everything. Not sure where we can put all of that 🤔

WDaan: I thought TRC was 'Torrent Right Click', so that can remain? 🤔 
Which would be the 'implemented version' of your generic right click component?

Larsluph: That's why I renamed it RightClickMenu ^^

WDaan: And the TRC menu got removed, so that one can be re-added? But using the generic component you made

Larsluph: It's already been extracted. Both the dashboard and content tab use the same RightClickMenu component.

WDaan: Yes, put everything TRC related can be moved back into the TRC component as it existed before? (just with changed syntax)
That way all that code isn't poluting the Dashboard component anymore.

Larsluph: Okay I got it, will do that

WDaan (approved):

PR #1475, opened 2024-01-20
chore(readme): Move installation to wiki

Larsluph:


WDaan (approved):

cybersholt:
Awesome idea, this'll help keep the readme clear. I was also wondering you're opinion of using something [docsify](https://github.com/docsifyjs/docsify) to create a dedicated documentation site like: https://jhildenbiddle.github.io/docsify-themeable/#/

Larsluph:
This may be a good idea but the wiki already uses Markdown files

I'm not sure I see the real added value here...

PR #1486, opened 2024-01-24
refactor on recent commits

Larsluph:


WDaan (approved):
Does this close delay work? 

Removing the 0 delay, solved the submenu bug

Larsluph:
> Does this close delay work?
> 
> Removing the 0 delay, solved the submenu bug

Yes but the 250ms delay is too much haha, I even tried with 1 and it still works

PR #1508, opened 2024-02-06
chore(localization): Automatically update config from Tolgee

Larsluph:


tolgee-action
WDaan, line 1: Why a submodule? Why not just use the workflow as a reusable workflow from another repository?

Larsluph: Because at first it wasn't a repo and the idea didn't come to mind 😅

WDaan (approved):

PR #1510, opened 2024-02-11
fix(QbitProvider): manually add indexes to getTorrentFiles response

alexisig:
# Manually add indexes to getTorrentFiles response

The absence of indexes in `getTorrentFiles`'s response made the request in `setTorrentFilePriority` impossible as it depends on them (id = index in the context).

The error only occurs with older version of Qbittorrent (< 4.4.0). PR that introduces indexes in the response: https://github.com/qbittorrent/qBittorrent/pull/14795

The fix manually add indexes on the returned array. 

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/services/QbitProvider.ts
+++ b/src/services/QbitProvider.ts
@@ -341,6 +341,16 @@ export default class QBitProvider implements IProvider {
         params: { hash, indexes: indexes?.join('|') }
       })
       .then(res => res.data)
+      .then(
+        (files: TorrentFile[]) => (files.some(file => file.index === undefined) ? files.map((file: TorrentFile, index: number) => ({ ...file, index })) : files)
+        /**
+         * We manually add indexes to the response if they are missing to provide compatibility with older versions of qbittorent (< 4.4.0)
+         * https://github.com/qbittorrent/qBittorrent/pull/14795
+         *
+         * We leave newer versions unaltered, as the files could be sent in different orders or be filtered
+         * https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#set-file-priority
+         */
+      )
   }
 
   async getAvailableTags(): Promise<string[]> {


Larsluph (approved):

WDaan (commented):
So you're overwriting the index if it were to exist before? 🤔

alexisig:
> So you're overwriting the index if it were to exist before? 🤔

Do you think it would be best to do it only if the indexes are missing / the version is < 4.4.0 first?

WDaan:
@Larsluph What do you think?

Only remap for the affected version and skip otherwise?

Larsluph:
Well it would be great but the service isn't aware of its own version so doing that might be a bit complicated for what it is.

Moreover files are sent in order so the indexes should match anyway

src/services/QbitProvider.ts
```diff
@@ -340,7 +340,14 @@ export default class QBitProvider implements IProvider {
       .get('/torrents/files', {
         params: { hash, indexes: indexes?.join('|') }
       })
-      .then(res => res.data)
+      .then(
+        res => res.data.map((file: TorrentFile, index: number) => ({ ...file, index }))
```
a65162, line 344: It is unnecessary to shallow copy for TorrentFile object.

alexisig:
I just checked [Qbitorrent's API documentation](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#set-file-priority) and it says this about the `filePrio`  endpoint :

> Since 2.8.2 it is reccomended to use index field returned by [torrent contents API](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#get-torrent-contents) (since the files can be filtered and the index value may differ from the position inside the response array).

2.8.2 API version == 4.4.0 Qbittorrent Version

----

How about something like this instead?
```javascript
  .then(res => res.data)
  .then((files: TorrentFile[]) =>
    files.some(file => file.index === undefined) ? 
      files.map((file: TorrentFile, index: number) => ({ ...file, index })) :
      files
  )
```

WDaan (approved):

PR #1531, opened 2024-02-20
fix: auto-reload when disabling VueTorrent

Larsluph:


WDaan (approved):

src/components/Dialogs/ImportSettingsDialog.vue
```diff
@@ -37,7 +37,7 @@ async function submit() {
   if (!isFormValid.value) return
 
   window.localStorage.setItem('vuetorrent_webuiSettings', settings.value)
-  window.location.reload()
+  location.reload()
```
WDaan, line 40: You sure this is available in this context?

Larsluph: We already use that when importing / resetting settings. It's just to align to a single syntax

Larsluph: Both window and location are in the global context of the browser

PR #1537, opened 2024-02-21
fix: Refactor pluralization logic in i18n.ts for "ru"

Saph1s:
# Refactor pluralization logic in i18n.ts for "ru"

Fixed translation when only 2 options are available

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/plugins/i18n.ts
+++ b/src/plugins/i18n.ts
@@ -13,6 +13,9 @@ export default createI18n({
       }
       const teen = choice > 10 && choice < 20
       const endsWithOne = choice % 10 === 1
+      if (choicesLength == 2) {
+        return choice === 1 ? 0 : 1
+      }
       if (choicesLength < 4) {
         return !teen && endsWithOne ? 1 : 2
       }


WDaan (approved):

src/plugins/i18n.ts
```diff
@@ -13,6 +13,9 @@ export default createI18n({
       }
       const teen = choice > 10 && choice < 20
       const endsWithOne = choice % 10 === 1
+      if (choicesLength == 2) {
+        return choice === 1 ? 0 : 1
```
Larsluph, line 17: I would include 0 and negative values in other languages. Could that be working with ru grammar?

```suggestion
        return choice <= 1 ? 0 : 1
```

Saph1s: Can you give an example?

Larsluph: Well for example in French and English, plural is applied when the associated numeric value is greater than 1, so all negative numbers, 0, 1, 0.5, ... are singular and 2, 1.1, 10000 are plural

Larsluph: After checking, this only applies in French. In English the 0 value is also plural.

Larsluph (approved):

PR #1547, opened 2024-02-26
feat: Add optional backend handling

Larsluph:
This PR implements a simple backend that stores configuration server-side.

https://github.com/VueTorrent/vuetorrent-backend

This backend is purely optional.

A new text field has appeared in the settings to enable it.
image

This field should contain the baseURL used to connect to that backend e.g. `http://localhost:8081/` or `https://subdomain.example.com/qbit/`.

On save and after that on app startup, it will send a ping to that backend if configured. If it is successful, configuration will starts to load from the backend instead of the localStorage. Everything fetched from the backend will be backed up to the localStorage for "offline" access.


Larsluph:
Race condition occurs when loading the app with an already configured backend. `useStorageAsync` calls happens before the initialization, skipping the first synchronization with the backend.

WDaan:
Pretty cool idea, that would allow people to sync settings across devices. Bit worried about the additional complexity this might generate; people having settings issues, people not understanding this setup and they have to host it.

What about authentication? How will this service be authenticated?

Larsluph:
> Pretty cool idea, that would allow people to sync settings across devices. Bit worried about the additional complexity this might generate;

Everything's handled by the persistance plugin for that so it shouldn't be a problem.

> people having settings issues, people not understanding this setup and they have to host it.

That's the main reason why I want to mark the fact that it is purely optional and the WebUI is still designed to be used alone. This backend will only "supercharge" the current UX.

> What about authentication? How will this service be authenticated?

Nothing's authenticated for now. I thought of how I could achieve that but the only possible way of doing it is using qbit's auth system as we can't really authenticate through it in the UI.

This solution did not seem really great so I kind of skipped that part assuming that the majority of users would either access their qbit instance through a "safe" channel (direct remote VPN) or by using an already authenticated way (I know some uses reverse proxy and stuff to automatically authenticate through qbittorrent).

Apart from that, I don't think not being authenticated is a major security flaw as the only thing "bad" you can do is delete someone's configuration.

I'm open if you have any suggestions about that, unfortunately I did not find a way to make it work flawlessly.

Larsluph:
Now that I think of it, the qbit sid cookie value should be sent to the backend in requests. It might be possible to check that against qbit directly to check the auth status 🤔

WDaan:
I think that could work 🤔

Larsluph:
@WDaan Tested on my home setup. Everything works with authentication! 🚀

WDaan:
Awesome! Sorry, lost track of this one.

It's seem like it's ready to go? Do we have kind of documentation in place to explain this new feature? Maybe the tooltip on the settings form can link back to a section in the Wiki or something?

Larsluph:
I updated the repo `docker-compose.yml` file to include it and some info are in the backend repo's readme.

I can add a Wiki entry to be more exhaustive on that. Not sure if it's possible to add a link in the tooltip but I can try.

WDaan (approved):

PR #1590, opened 2024-03-19
perf: add title navigation

cameronaw13:
# Add title navigation

Not sure if I can speak for everyone, but I found it annoying that clicking the title while viewing torrent details or webui settings doesn't route you back to the homescreen. This pr fixes that.

If anyone has any issues with the implementation/styling of this, please let me know :)

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Navbar/Navbar.vue
+++ b/src/components/Navbar/Navbar.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { useDashboardStore, useNavbarStore, useVueTorrentStore } from '@/stores'
 import { storeToRefs } from 'pinia'
+import { useRouter } from 'vue-router'
 import BottomActions from './SideWidgets/BottomActions.vue'
 import CurrentSpeed from './SideWidgets/CurrentSpeed.vue'
 import FilterSelect from './SideWidgets/FilterSelect.vue'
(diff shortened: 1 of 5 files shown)


src/components/Navbar/Navbar.vue
```diff
@@ -60,8 +66,10 @@ const toggleDrawer = () => {
   <v-app-bar class="ios-padding">
     <v-app-bar-nav-icon @click="toggleDrawer" />
     <v-app-bar-title>
-      <span v-if="$vuetify.display.smAndUp" class="text-accent">Vue</span>
-      <span v-if="$vuetify.display.smAndUp">Torrent</span>
+      <v-app-bar-title style="display: inline-flex; width: min-content; cursor: pointer" @click="goHome">
```
Larsluph, line 69: Can you explain the inline styles?

cameronaw13: Display: inline-flex allows the active filters div to sit side by side with the app-bar title since its now technically two separate divs.

Width: min-content resizes the inner v-app-bar-title to be the same size as the title representing the area you can click to return to the homepage. The reason why I created a wrapper element for the title is because v-app-bar-title spans the width of the entire app-bar and cannot be changed without also changing the top-actions' positioning.

Cursor: pointer make the cursor a pointer so that users know that you can click on the title to return to the homepage.

If you'd prefer this within the styles.scss file I can make that change.

Larsluph: Apart from that it's great!

Larsluph: No need to declare that class globally, declare it inside the component as it will only be used there.

cameronaw13: Ah that makes more sense, should be fixed now.

Larsluph (approved):

cameronaw13:
Sorry, thought this was done but the lack of padding for the clickable area really bugged me. Should be ready to merge if all is well.

Larsluph (approved):
New margins are perfect!

WDaan (approved):

PR #1611, opened 2024-03-25
perf: Replace ratio_time_limit by seeding_time_limit

Larsluph:
⚠️ This may cause undefined keys in the properties configuration. Reset your settings to have the latest schema.


WDaan:
Why this change? 🤔

WDaan (approved):

Issue #895, opened 2023-06-23
Add bulk rename feature


ticklemyIP:
This is something this project could be unique in. There's currently no known way (by me) of bulk renaming torrents' files "in house" (through the torrent software). Why this would be preferred as opposed to regular renaming, is a user wouldn't always be able to (or want to) ssh into a box or (in a better situation) open a file manager, or use a different crappy bulk renamer. And also this way the torrent sooftware wouldn't find the new and renamed file wich would disable seeding or downloading.

This feature would be most useful with mediaservers which often use a very strict naming scheme in order to recognize media and apply metadata, which torrent uploaders doesn't always follow. This is particularly true with shows, thus bulk renaming would be a much better way of managing these files.

Larsluph:
Duplicate of #752 ?

ticklemyIP:
Could be, but I think my issue is a bit more detailed and gives better reasons. Up to you which one you want to keep of course.

gpollo:
In [killemov/Shift](https://github.com/killemov/Shift) custom UI for Transmission, the batch rename simply uses Regex patterns.

A UI as simple this could work very well.

```
+--------------------------------------+
|                                      |
| Pattern:     $tv.show.(S..E..)*.mkv^ |
|                                      |
| Replacement: TV Show - \1.mkv        |
|                                      |
|                                Apply |
+--------------------------------------+
```

If I have time, I may have a look at implementing that.

loligans:
I also implemented this functionality in the main qBittorrent WebUI here: https://github.com/qbittorrent/qBittorrent/pull/18287

Larsluph:
Fixed with 1624

PR #1624, opened 2024-03-29
feat(TorrentDetail): add bulk renaming

yunyuyuan:
# feat(TorrentDetail): add bulk renaming

https://github.com/VueTorrent/VueTorrent/assets/45785585/af75ef38-4606-4406-be48-e8c02ea66686

For #895.

The main functions are primarily referenced from https://github.com/qbittorrent/qBittorrent/pull/18287.

**Features**:
- Using regular expressions to handle renaming, supporting flags
- Expand/collapse folders, partial selection
- Identifying duplicate file names
- Identifying renaming result conflicts with original name

**Limitations**:
- Unable to rename folders
- Regular expressions only apply to file name and cannot be applied to the full path

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Core/HistoryField.vue
+++ b/src/components/Core/HistoryField.vue
@@ -1,13 +1,11 @@
 <script setup lang="ts">
 import { HistoryKey } from '@/constants/vuetorrent'
 import { useHistoryStore } from '@/stores'
-import { computed, ref, useSlots } from 'vue'
+import { computed, ref } from 'vue'
 
 const props = defineProps<{
   historyKey: HistoryKey
 }>()
-
-const slots = useSlots()
 
 const _value = defineModel<string | undefined>({ required: true })
 
@@ -29,7 +27,7 @@ defineExpose({
 
 <template>
   <v-combobox v-model="_value" ref="field" :items="historyValue">
-    <template v-slot:prepend v-if="slots.prepend"><slot name="prepend" /></template>
+    <template v-slot:prepend v-if="$slots.prepend"><slot name="prepend" /></template>
   </v-combobox>
 </template>
 
--- a/src/components/Core/HistoryField.vue
+++ b/src/components/Core/HistoryField.vue
@@ -1,11 +1,13 @@
 <script setup lang="ts">
 import { HistoryKey } from '@/constants/vuetorrent'
 import { useHistoryStore } from '@/stores'
-import { computed, ref } from 'vue'
+import { computed, ref, useSlots } from 'vue'
 
 const props = defineProps<{
   historyKey: HistoryKey
 }>()
+
+const slots = useSlots()
 
 const _value = defineModel<string | undefined>({ required: true })
 
@@ -27,7 +29,7 @@ defineExpose({
 
 <template>
   <v-combobox v-model="_value" ref="field" :items="historyValue">
-    <template v-slot:prepend><slot name="prepend" /></template>
+    <template v-slot:prepend v-if="slots.prepend"><slot name="prepend" /></template>
   </v-combobox>
 </template>
 
--- a/src/components/Dialogs/BulkRenameFilesDialog.vue
+++ b/src/components/Dialogs/BulkRenameFilesDialog.vue
@@ -157,7 +157,11 @@ const fileCheckChange = (item: ItemRow) => {
   dryRunRename([item])
 }
 
-const dryRunRename = (partialItems?: ItemRow[]) => {
+const dryRunRename = async (partialItems?: ItemRow[]) => {
+  await form.value?.validate()
+  if (!isFormValid.value) {
+    return
+  }
   let regexp: RegExp
   try {
     regexp = new RegExp(regexpInput.value, regexpFlagsInput.value.join(''))
@@ -166,7 +170,7 @@ const dryRunRename = (partialItems?: ItemRow[]) => {
   }
   ;(partialItems ? partialItems : items).forEach(item => {
     if (item.type === 'file') {
-      if (isFormValid.value && item.selected && regexp.test(item.name)) {
+      if (item.selected && regexp.test(item.name)) {
         item.targetName = item.name.replace(regexp, targetInput.value)
         item.targetFullName = (item.parentItem!.fullName === '' ? '' : item.parentItem!.fullName + '/') + item.targetName
       } else {
@@ -241,7 +245,7 @@ onMounted(() => {
       <v-card-text class="d-flex flex-column">
         <v-form v-model="isFormValid" ref="form">
           <v-row no-gutters align="center" justify="center">
-            <v-col :cols="$vuetify.display.mobile ? 12 : undefined">
+            <v-col :cols="$vuetify.display.mobile ? 9 : undefined">
               <HistoryField
                 :historyKey="HistoryKey.BULK_RENAME_REGEXP"
                 ref="regexpEl"
@@ -250,17 +254,18 @@ onMounted(() => {
                 v-model="regexpInput"
                 :rules="rules"
                 :label="$t('dialogs.bulkRenameFiles.regexp')">
-                <template v-slot:append>
-                  <v-select
-                    v-model="regexpFlagsInput"
-                    :items="['d', 'g', 'i', 'm', 's', 'u', 'v', 'y']"
-                    :placeholder="t('dialogs.bulkRenameFiles.select_regex_flags')"
-                    label="Flags"
-                    density="compact"
-                    multiple
-                    hide-details></v-select>
-                </template>
               </HistoryField>
+            </v-col>
+            <v-col :cols="$vuetify.display.mobile ? 3 : 'auto'">
+              <v-select
+                class="ml-2"
+                v-model="regexpFlagsInput"
+                :items="['d', 'g', 'i', 'm', 's', 'u', 'v', 'y']"
+                :placeholder="t('dialogs.bulkRenameFiles.select_regex_flags')"
+                label="Flags"
+                density="compact"
+                multiple
+                hide-details></v-select>
             </v-col>
             <v-col cols="auto">
               <v-icon class="mx-2" :icon="`mdi-arrow-${$vuetify.display.mobile ? 'down' : 'right'}`" />
@@ -273,7 +278,8 @@ onMounted(() => {
                 density="compact"
                 v-model="targetInput"
                 :rules="rules"
-                :label="$t('dialogs.bulkRenameFiles.target')" />
+                :label="$t('dialogs.bulkRenameFiles.target')">
+              </HistoryField>
             </v-col>
             <v-col cols="auto">
               <v-badge :class="$vuetify.display.mobile ? 'mt-2' : 'ml-5'" color="success" location="top left" :content="candidateItems.length">
@@ -330,9 +336,6 @@ onMounted(() => {
 </template>
 
 <style lang="scss" scoped>
-.v-select {
-  min-width: 100px;
-}
 .v-card-text {
   height: calc(100vh - 112px);
 }
--- a/src/components/Dialogs/BulkRenameFilesDialog.vue
+++ b/src/components/Dialogs/BulkRenameFilesDialog.vue
@@ -1,9 +1,10 @@
 <script setup lang="ts">
 import { useDialog } from '@/composables'
+import HistoryField from '@/components/Core/HistoryField.vue'
 import { getFileIcon } from '@/constants/vuetorrent'
 import { useContentStore } from '@/stores'
 import { TreeFolder, TreeNode } from '@/types/vuetorrent'
-import { storeToRefs } from 'pinia'
+import { HistoryKey } from '@/constants/vuetorrent'
 import { reactive, ref, onMounted, watch, computed } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { toast } from 'vue3-toastify'
@@ -22,11 +23,11 @@ const contentStore = useContentStore()
 const form = ref<VForm>()
 const isFormValid = ref(false)
 const hasDuplicated = ref(false)
-const {
-  bulkRenameRegexp: regexpInput,
-  bulkRenameRegexpFlags: regexpFlagsInput,
-  bulkRenameTarget: targetInput
-} = storeToRefs(contentStore)
+const regexpInput = ref('')
+const regexpEl = ref<typeof HistoryField>()
+const regexpFlagsInput = ref([])
+const targetInput = ref('')
+const targetEl = ref<typeof HistoryField>()
 const running = ref(false)
 
 const rules = [(v: string) => !!v]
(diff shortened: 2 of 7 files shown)


loligans:
Glad to see this feature being implemented!

loligans:
Can you explain the limitations a little more? Is it limited by functionality of the WebApp or due to the complexity of implementing said functionality

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,317 @@
...
+                hide-details></v-select>
+            </template>
+          </v-text-field>
+          ->
+          <v-text-field hide-details density="compact" v-model="targetInput" :rules="rules" :label="$t('dialogs.bulkRenameFiles.target')" />
+          <v-badge color="success" :content="candidateItems.length">
+            <v-btn :loading="running" :disabled="!isFormValid || hasDuplicated" color="primary" @click="run()">{{ $t('dialogs.bulkRenameFiles.run') }}</v-btn>
+          </v-badge>
+        </v-form>
+        <v-data-table-virtual :headers="headers" :items="items" density="compact" height="calc(100vh - 200px)" fixed-header>
```
Larsluph, line 253: UI not responsive at all, please rework by using virtual scroller instead of virtual data table.

See Content view for example.

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,317 @@
...
+              <template v-slot:item.selected>
+                <v-checkbox-btn v-if="item.type === 'file'" v-model="item.selected" :color="item.targetName && 'indigo'" @change="fileCheckChange(item)" />
+                <v-checkbox-btn v-else v-model="item.selected" :indeterminate="item.indeterminate" @change="folderCheckChange(item)" />
+              </template>
+              <template v-slot:item.name>
+                <span class="fold-toggle" :class="{ clickable: item.type === 'folder' }" @click="item.type === 'folder' && toggleFolderFolded(item, !item.folded)">
+                  <v-tooltip v-if="item.type === 'folder'" location="top" activator="parent">
+                    {{ t(`dialogs.bulkRenameFiles.${item.folded ? 'unfold' : 'fold'}`) }}
+                  </v-tooltip>
+                  {{ '&emsp;'.repeat(item.indent) }}
```
Larsluph, line 265: Is there a reason for using HTML space entities instead of classical margin / padding?

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,317 @@
...
+                  <v-icon v-if="item.type === 'folder'">{{ item.folded ? 'mdi-chevron-down' : 'mdi-chevron-up' }}</v-icon>
+                  <v-icon v-if="node.fullName === ''" icon="mdi-file-tree" />
```
Larsluph, line 267: When opening the dialog through the root element, all icons are `mdi-file-tree`. It works properly when opening it through a folder.

```suggestion
                  <v-icon v-if="item.fullName === ''" icon="mdi-file-tree" />
```

src/stores/vuetorrent.ts
```diff
@@ -35,6 +35,9 @@ export const useVueTorrentStore = defineStore(
     const refreshInterval = ref(2000)
     const fileContentInterval = ref(5000)
     const useIdForRssLinks = ref(false)
+    const bulkRenameRegexp = ref('')
+    const bulkRenameRegexpFlags = ref([])
+    const bulkRenameTarget = ref('')
```
Larsluph, line 40: Put it in the content store instead

Larsluph (changes requested):

Larsluph:
Lot of things to say on this PR, great feature to tackle anyway.

This still misses some work to allow for proper UX, only tried for a few minutes and already saw some errors (tree icon not matching, regex flag throwing errors, feature not usable on mobile and not practical on wider screens). Try thinking vertically instead of horizontally for better responsiveness.

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,317 @@
...
+  dryRunRename()
+})
+
+onMounted(() => {
+  parseNode(props.node)
+})
+</script>
+
+<template>
+  <v-dialog v-model="isOpened" persistent>
```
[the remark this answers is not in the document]
yunyuyuan: Bulk renaming can indeed be a complex operation, and users may unexpectedlly close the dialog box by click either `esc` or mask area.

yunyuyuan:
@loligans 

The inability to rename folders is due to the potential for altering the file tree, leading to unexpected results. 

Regular expressions only apply to names rather than the entire path, as modifying the file tree, for instance, renaming `Season 1/xxx.mkv` to `Seasons/01/xxx.mkv`, could occur and it's a bit hard to handle. 

Additionally, renaming folders is a separate function. It might be worth considering implementing a dialog box for batch renaming folders only.

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,317 @@
...
+  dryRunRename()
+}
+
+const fileCheckChange = (item: ItemRow) => {
+  detectIndeterminate(props.node)
+  dryRunRename([item])
+}
+
+const dryRunRename = (partialItems?: ItemRow[]) => {
+  const regexp = new RegExp(regexpInput.value, regexpFlagsInput.value.join(''))
```
yunyuyuan, line 159:
```suggestion
let regexp: RegExp;
try {
  regexp = new RegExp(regexpInput.value, regexpFlagsInput.value.join(''))
} catch {
  return
}
```

@Larsluph Regarding "regex flag throwing errors," do you mean changing it like this?

src/components/Dialogs/BulkRenameFilesDialog.vue
yunyuyuan: I haven't found a good solution. Vertical display might be tricky to handle. Currently, I have it set to not wrap, allowing horizontal scrolling.

image

loligans:
That makes sense. There is a way around enabling folders to be renamed without indeterminate results. Idk how different the data is represented in the Vue app, but the way I did it was rename in reverse order so that all files under a folder are renamed before the folder is renamed. This way allows you to rename files and folders with the same logic. Then I included a toggle to enable/disable folders in the rename logic.

yunyuyuan:
@loligans  I've implemented nested selection logic for checkboxes in folders, ensuring that selecting a folder also selects its internal files. (This is because I found it cumbersome to use your webUI dialog without being able to batch select a folder🤣)

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
Larsluph: If you don't handle closing the dialog as part of the process, then why include refs in the store instead of directly in the dialog?

src/components/Dialogs/BulkRenameFilesDialog.vue
Larsluph: Can you try putting the diff on 2 lines instead of next to each other?

I think of putting files without change one line like the current layout but changing files would display on two lines, with the top one being the "old" one, highlighted in red or striked-through. And the "new" one being at the bottom and green?

src/components/Dialogs/BulkRenameFilesDialog.vue
Larsluph: It would prevent basic error for people that don't know these flags.
This looks like a great quick fix.

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
yunyuyuan: for reusing the regexpinput and replacementinput

src/components/Dialogs/BulkRenameFilesDialog.vue
yunyuyuan: I don't know which one is better,imo I like one line display😂, two lines in one <td> maybe too compact.

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
Larsluph: In that case why not include the selection as well? You can set a `v-model` on the data table to bind it.

src/components/Dialogs/BulkRenameFilesDialog.vue
Larsluph: That's why I suggested using virtual scroller instead of using the data table 😄

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
yunyuyuan: who will rename one file many times😏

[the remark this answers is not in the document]
Larsluph: But the content store will be cleared everytime you close the torrent detail page. Storing the selection with the other fields will allow to close the dialog without loosing changes.

If you want to keep field values for input and replacement, I would suggest you use a `HistoryField` instead.

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,336 @@
...
+  folded?: boolean // folder only
+  indeterminate?: boolean // folder only
+  parentItem?: ItemRow
+  duplicated?: boolean
+  notChanged?: boolean
+  targetName?: string
+  targetFullName?: string
+  node: T
+} & (T extends TreeFolder ? { type: 'folder' } : { type: 'file' })
+const items = reactive<ItemRow[]>([])
```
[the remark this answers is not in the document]
yunyuyuan: because it's reactive?

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
yunyuyuan: get it

src/components/Core/HistoryField.vue
```diff
@@ -27,7 +29,7 @@ defineExpose({
 
 <template>
   <v-combobox v-model="_value" ref="field" :items="historyValue">
-    <template v-slot:prepend><slot name="prepend" /></template>
+    <template v-slot:prepend v-if="slots.prepend"><slot name="prepend" /></template>
```
Larsluph, line 32: use `$slots` instead of importing `useSlots`

src/components/Dialogs/BulkRenameFilesDialog.vue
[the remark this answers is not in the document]
Larsluph: I wasn't aware about its use on arrays but it looks fine, you can keep it like that.

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,361 @@
...
+</script>
+
+<template>
+  <v-dialog v-model="isOpened" persistent>
+    <v-card density="compact">
+      <v-toolbar color="transparent">
+        <v-toolbar-title>{{ $t('dialogs.bulkRenameFiles.title') }}</v-toolbar-title>
+        <v-btn icon="mdi-close" @click="close()" />
+      </v-toolbar>
+      <v-card-text class="d-flex flex-column">
```
Larsluph, line 245: Missing `v-card-title` element wiht ios safe margin
```suggestion
    <v-card density="compact">
      <v-card-title class="ios-margin">
        <v-toolbar color="transparent">
          <v-toolbar-title>{{ $t('dialogs.bulkRenameFiles.title') }}</v-toolbar-title>
          <v-btn icon="mdi-close" @click="close()" />
        </v-toolbar>
      </v-card-title>
      <v-card-text class="d-flex flex-column">
```

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,361 @@
...
+            <v-col :cols="$vuetify.display.mobile ? 9 : undefined">
+              <HistoryField
+                :historyKey="HistoryKey.BULK_RENAME_REGEXP"
+                ref="regexpEl"
+                hide-details
+                density="compact"
+                v-model="regexpInput"
+                :rules="rules"
+                :label="$t('dialogs.bulkRenameFiles.regexp')">
+              </HistoryField>
```
Larsluph, line 257:
```suggestion
                :label="$t('dialogs.bulkRenameFiles.regexp')" />
```

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,361 @@
...
+            <v-col :cols="$vuetify.display.mobile ? 3 : 'auto'">
+              <v-select
+                class="ml-2"
+                v-model="regexpFlagsInput"
+                :items="['d', 'g', 'i', 'm', 's', 'u', 'v', 'y']"
+                :placeholder="t('dialogs.bulkRenameFiles.select_regex_flags')"
+                label="Flags"
+                density="compact"
+                multiple
+                hide-details></v-select>
```
Larsluph, line 268:
```suggestion
                hide-details />
```

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,361 @@
...
+            <v-col :cols="$vuetify.display.mobile ? 12 : undefined">
+              <HistoryField
+                :historyKey="HistoryKey.BULK_RENAME_TARGET"
+                ref="targetEl"
+                hide-details
+                density="compact"
+                v-model="targetInput"
+                :rules="rules"
+                :label="$t('dialogs.bulkRenameFiles.target')">
+              </HistoryField>
```
Larsluph, line 282:
```suggestion
                :label="$t('dialogs.bulkRenameFiles.target')" />
```

src/components/Dialogs/BulkRenameFilesDialog.vue
```diff
@@ -0,0 +1,361 @@
...
+                  <v-icon v-if="item.fullName === ''" icon="mdi-file-tree" />
+                  <v-icon v-else-if="item.type === 'file'" :icon="getFileIcon(item.name)" />
+                  <v-icon v-else-if="!item.folded" icon="mdi-folder-open" color="#ffe476" />
+                  <v-icon v-else icon="mdi-folder" color="#ffe476" />
+                  {{ item.name }}
+                </span>
+              </template>
+              <template v-slot:item.targetName>
+                <span v-if="item.type === 'file'" class="target-name" :class="{ duplicated: item.duplicated, 'not-changed': item.notChanged }"
+                  >{{ item.targetName }}
```
Larsluph, line 317:
```suggestion
                <span v-if="item.type === 'file'" class="target-name" :class="{ duplicated: item.duplicated, 'not-changed': item.notChanged }">
                  {{ item.targetName }}
```

src/components/Dialogs/BulkRenameFilesDialog.vue
yunyuyuan: What about this?

https://github.com/VueTorrent/VueTorrent/assets/45785585/2ba72c12-1a6b-439b-92f5-8167c77c4742

Larsluph (approved):
Tested with 865 files and it's a bit laggy, virtual table is used so I'll pass on that.
Good job on that feature, it'll be a great addition 🚀

yunyuyuan:
@Larsluph  
Yes, there are significant performance issues. Now, I've rewritten part of the code, and when testing with 960 items, the delay for selecting all is approximately 2.5 seconds, compared to around 5 seconds before optimization.

WDaan (approved):

yunyuyuan:
> Then I included a toggle to enable/disable folders in the rename logic.

@loligans I'm really sorry about that I said `This is because I found it cumbersome to use your webUI dialog without being able to batch select a folder🤣`!!! I didn't notice that sentence above, it turns out you provided a right-click menu for Toggle Selection! In your solution, renaming folders and batch selecting/deselecting can be perfectly achieved. It's indeed my mistake.

image

PR #1641, opened 2024-04-13
perf(colors): Rework color palette to better match accent color

### LINKED ISSUE #1639: [Feature Request]: Darken drawer background
### Description

First all thank you for an fantastic skin for qbittorent! The mobile view is especially a massive help for daily use.

However, the drawer background color is **driving me nuts**. It is too close to the "highlighted element" color most browsers devtools use as well as being close to "selected content by dragging the mouse across a page" highlighter color.

image

My monkey brain sees the drawer color and thinks "this is erroneously highlighted for some reason" and just can't ignore it. :skull: 

Unfortunately I can't build my own VueTorrent because I run qbittorent through a container and use lsio docker mods for vuetorrent.

**I know this is quite a selfish and personal request but would it be possible to darken this color enough that it no longer resembles that highlighted color?**

Something like `rgb(26, 33, 40)` would be enough

image

Currently I am using tampermonkey to modify styling on load

```js
function waitForElm(selector) {
    return new Promise(resolve => {
        if (document.querySelector(selector)) {
            return resolve(document.querySelector(selector));
        }

        const observer = new MutationObserver(mutations => {
            if (document.querySelector(selector)) {
                observer.disconnect();
                resolve(document.querySelector(selector));
            }
        });
        observer.observe(document.body, {
            childList: true,
            subtree: true
        });
    });
}

(function() {
    waitForElm('nav.v-navigation-drawer').then((elm) => {
        console.log('Element is ready');
        document.querySelectorAll('nav.v-navigation-drawer .bg-primary').forEach(elm => elm.classList.remove('bg-primary'));
        document.querySelector('nav.v-navigation-drawer').style.setProperty('background-color', 'rgb(26, 33, 40)', 'important');
    });
})();
```

But this isn't feasible for mobile. And I'd prefer a true fix rather than hacking it like this.

Thanks for your time!

Larsluph:
Navbar color has been darkened
image

Also added a redesigned variant to better match the app's visual identity.
image

With all that, light and dark mode now has an option to specify which variant you want
image

Colors handling also moved to plain TS for better support of theme in the future as it will allow for fully customizable theme colors.

BTW torrent card sideborders are fixed in that PR.

Fixes #1639


Larsluph:
> Seems like this issue is not being super upvoted by everybody, so perhaps this should be more of a public vote?

Totally agree with that, no idea how we could achieve that? In the discord maybe? I heard they are currently releasing a polling system, we could try it.

> We could also make it customizable but I'm a bit afraid of the tech debt overhead that could/would generate.

Adding several color palettes in the interface could be possible thanks to vuetify's theme customization. I don't think it's possible to make it customizable though...

Larsluph:
@WDaan for this would you prefer a toggle to enable these alternative colors, or making a poll to replace the existing palette?

WDaan:
I guess we could use Discord for some community input? What would they like to see changed? Maybe some minor adjustments are sufficient to resolve the issue. Nobody else has commented on it or posted a reaction.

WDaan:
Allright guess both will exist, and they will be slightly tweaked over time

Larsluph:
> Personally, I like the green on the old version of the chart more.

The green on the right is the correct one for the legacy theme.
The green on the left is for the redesigned theme. I might have copy-pasted too quickly or I didn't understand what you said 🤔

> I agree the background of the sidebar could be a little darker but this current tint is a bit too dark for my taste. Or perhaps I'm just so used of seeing the old theme 😅

Totally agree but this dark tint is for fixing the initial issue, unless we keep it the same as before if we assume that the issue isn't relevant...

> The redesigned theme is cool, but I fail to see how it's easier to read.

Redesigned theme was just something I had in mind to keep a monochromatic color scheme throughout the UI. It doesn't have any added value.

> Also I think the legacy theme should remain the default? 🤔 

Forgot to change if back after debugging 😅

WDaan:
Screenshot 2024-05-29 at 19 24 03

In the new version of the old theme, the green in the chart is also updated. So now it no longer matches the other green from the cards as it did before. Not sure if we're trying to keep them similar or not.

WDaan (approved):
LGTM 🚀 great!

PR #1690, opened 2024-05-17
feat(pages): Deploy mocked demo version

Larsluph:
See temporary preview here: https://larsluph.github.io/demo/
Will be moved to vuetorrent subdomain after merge: https://github.com/VueTorrent/demo

Had to remove tree-shaking made for the MockProvider as it would interfere with the app render and completely break the demo page. Build size `+~500kb`


WDaan (approved):
This is super cool 🚀 Can't believe we didn't do this sooner 😄

PR #1701, opened 2024-05-23
feat(tags): added remove all button for tags in rightClick menu

### LINKED ISSUE #1670: [Feature Request]: Add "remove all tags" action in right click menu
### Description

in the classic qbit webui under tags there is an option to remove all tags from a torrent. This is handy when manually cleaning up after qbitmanage or wanting to reset tags in general

vuetorrent only allows to deselect each tag one by one with a delay between each.

Add "Remove all tags" at the top of the tag right click menu.

switch to classic UI and do the same

illeb:
## added remove all button for tags in rightClick menu [feat]

Fixes #1670

Screenshot 2024-05-23 alle 23 38 29

Clicking the "Remove all" button removes all the tags from a particular torrent.

Obviously the element is visible only if tags are present.

NOTE:
I had to modify the multiple tags separator from `removeTorrentTag` in `src/services/qbit/QbitProvider.ts` because the separator used was wrong (it was using `|` but i think the correct one is the `,`)
see more in the example of the api specification: https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#delete-tags
## PR Checklist

- [x] All tests pass


src/components/Dashboard/RightClick.vue
```diff
@@ -207,11 +211,22 @@ const menuData = computed<RightClickMenuEntryType[]>(() => [
...
-      text: tag,
-      icon: hasTag(tag) ? 'mdi-checkbox-marked' : 'mdi-checkbox-blank-outline',
-      action: async () => await toggleTag(tag)
-    }))
+    children: [
+      ...(torrent.value?.tags.length
+        ? [
+            {
+              text: t('dashboard.right_click.tags.removeall'),
+              action: async () => await removeTags(maindataStore.tags),
```
Larsluph, line 219: qBittorrent will remove all tags from the torrents if none are specified in the request, i would use this rule instead to prevent bugs linked to local tags cache and to reduce request size.

src/components/Dashboard/RightClick.vue
```diff
@@ -207,11 +211,22 @@ const menuData = computed<RightClickMenuEntryType[]>(() => [
...
-    children: maindataStore.tags.map(tag => ({
-      text: tag,
-      icon: hasTag(tag) ? 'mdi-checkbox-marked' : 'mdi-checkbox-blank-outline',
-      action: async () => await toggleTag(tag)
-    }))
+    children: [
+      ...(torrent.value?.tags.length
+        ? [
+            {
+              text: t('dashboard.right_click.tags.removeall'),
```
Larsluph, line 218: Rename locale key to match format

```suggestion
              text: t('dashboard.right_click.tags.remove_all'),
```

src/components/Dashboard/RightClick.vue
```diff
@@ -83,6 +83,10 @@ function hasTag(tag: string) {
   return torrents.value.every(torrent => torrent && torrent.tags && torrent.tags.includes(tag))
 }
 
+async function removeTags(tags: string[]) {
```
Larsluph, line 86: This function can thus be renamed to `removeAllTags`

Larsluph (changes requested):
Great PR overall, thanks for the contribution!

src/components/Dashboard/RightClick.vue
```diff
@@ -83,8 +83,8 @@ function hasTag(tag: string) {
   return torrents.value.every(torrent => torrent && torrent.tags && torrent.tags.includes(tag))
 }
 
-async function removeTags(tags: string[]) {
-  await torrentStore.removeTorrentTags(hashes.value, tags)
+async function removeAllTags() {
+  await torrentStore.removeTorrentTags(hashes.value, [])
```
Larsluph, line 87: Passing an empty list won't trigger the effect, you will need to make that parameter optional all the way through the qbit provider (and its interface)

src/components/Dashboard/RightClick.vue
```diff
@@ -215,8 +215,8 @@ const menuData = computed<RightClickMenuEntryType[]>(() => [
       ...(torrent.value?.tags.length
         ? [
             {
-              text: t('dashboard.right_click.tags.removeall'),
-              action: async () => await removeTags(maindataStore.tags),
+              text: t('dashboard.right_click.tags.remove_all'),
+              action: async () => await removeAllTags(),
```
Larsluph, line 219: *(Optional)*: You can also pass the function directly without surrounding it in a lambda
```suggestion
              action: removeAllTags,
```

src/components/Dashboard/RightClick.vue
illeb: Thanks for the feedback, just commited a fix 👍

src/services/qbit/IProvider.ts
```diff
@@ -463,9 +463,9 @@ export default interface IProvider {
   /**
    * Remove tags to torrents
    * @param hashes Torrent hashes
-   * @param tags Tags to remove
+   * @param tags Tags to remove. If not provided, all tags will be removed
    */
-  removeTorrentTag(hashes: string[], tags: string[]): Promise<void>
+  removeTorrentTag(hashes: string[], tags?: string[]): Promise<void>
```
Larsluph, line 468: Also change `MockProvider` signature

src/services/qbit/IProvider.ts
```diff
@@ -463,9 +463,9 @@ export default interface IProvider {
   /**
    * Remove tags to torrents
    * @param hashes Torrent hashes
-   * @param tags Tags to remove
+   * @param tags Tags to remove. If not provided, all tags will be removed
```
Larsluph, line 466: 👍

src/services/qbit/QbitProvider.ts
```diff
@@ -563,8 +563,10 @@ export default class QBitProvider implements IProvider {
     return this.torrentAction('addTags', hashes, { tags: tags.join('|') })
   }
 
-  async removeTorrentTag(hashes: string[], tags: string[]): Promise<void> {
-    return this.torrentAction('removeTags', hashes, { tags: tags.join('|') })
+  async removeTorrentTag(hashes: string[], tags?: string[]): Promise<void> {
+    return tags
+      ? this.torrentAction('removeTags', hashes, { tags: tags.join(',') })
+      : this.torrentAction('removeTags', hashes)
```
Larsluph, line 569: Reduce duplicated code
```suggestion
    let options = tags ? { tags: tags.join(',') } : undefined
    return this.torrentAction('removeTags', hashes, options)
```

WDaan (approved):

Larsluph:
Tolgee project has been updated 👍

Issue #1492, opened 2024-01-30
[Feature Request]: option to have tags/categories without the “pill” shape


nickelslol:
### Description

I would like an option to not have the coloured pill shapes around tags and categories

### Describe the solution you'd like

Just have plain text

I do see the benefit of having the colours so would suggest the text gets coloured in the same colour tags and categories are now 

### Complementary informations

- [ ] Is this feature already implemented in the default WebUI?

PR #1708, opened 2024-05-28
feat(tags): option to have tags/categories without the “pill” shape (#1492)

jagadam97:
# feat(tags): option to have tags/categories without the “pill” shape (#1492)

- added an option to configure tags/categories without pill shape.
- can be achived thourgh enabling hide colered chips option in general settings.
image
image
image
image
image
https://youtu.be/42Y7L2RvH2g

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -971,6 +971,7 @@
         "enableRatioColors": "Enable ratio colors",
         "fileContentInterval": "Torrent file content refresh interval",
         "hideChipIfUnset": "Hide chips if unset",
+        "hideColoredChip": "Use text variant for chips",
         "historySize": "History size on eligible fields",
         "import": "Import Settings",
         "isDrawerRight": "Right Drawer",
--- a/src/components/Core/ColoredChip.vue
+++ b/src/components/Core/ColoredChip.vue
@@ -18,14 +18,15 @@ const props = withDefaults(
 )
 
 const { t } = useI18n()
-const { enableHashColors } = storeToRefs(useVueTorrentStore())
+const { enableHashColors, hideColoredChip } = storeToRefs(useVueTorrentStore())
 
 const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value)))
 const chipValue = computed(() => (props.disabled ? props.disabledValue || t('common.none') : props.value))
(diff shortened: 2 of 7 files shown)


src/locales/en.json
```diff
@@ -960,6 +960,7 @@
         "enableRatioColors": "Enable ratio colors",
         "fileContentInterval": "Torrent file content refresh interval",
         "hideChipIfUnset": "Hide chips if unset",
+        "hideColoredChip": "Hide colored chips (Text with color)",
```
Larsluph, line 963: I would opt for something more explicit like
```suggestion
        "useTextForChips": "Use text variant for chips",
```

Larsluph (commented):
Also I think you forgot to change the chips inside the Info tab, at the bottom you may see tags listed using chips.

src/locales/en.json
jagadam97: > I would opt for something more explicit like

these format seems better. I will get this implemented

jagadam97:
> Also I think you forgot to change the chips inside the Info tab, at the bottom you may see tags listed using chips.

yep forgot about that page... will get that screen covered too

WDaan:
Does this change need to be optional? 🤔 

Can't we just not show anything when there is no tag/category? 
I don't think anybody will be opposed and supporting more and more customizable options also creates additional overhead.

Larsluph (approved):

nickelslol:
Hey, i really appreciate someone taking the time to implement something i asked for. Would you be able to take away what I assume is padding around the text, at the moment it’s like just an invisible pill

nickelslol:
Just to clarify my reasoning. The text doesn't line up with other text on the same line the way it is now, and taking up less space would allow more to be shown on smaller screens

Larsluph:
@nickelslol I pushed a fix to address that: #1712

jagadam97:
> Hey, i really appreciate someone taking the time to implement something i asked for. Would you be able to take away what I assume is padding around the text, at the moment it’s like just an invisible pill

i too had the same concern about the padding and all, but did not wanted to change the component too much unless required (if any one else got bothered). I see that its resolved very quickly. thank for taking time to check this feature out @nickelslol. thanks for fixing it @Larsluph

PR #1751, opened 2024-06-24
chore: Add mail contact for private requests

Larsluph:
Add mail option to issue creation page for private requests and users without Discord accounts (either due to geo bans or for personal reasons)

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -4,5 +4,5 @@ contact_links:
     url: https://discord.gg/KDQP7fR467
     about: Join the discord server to contact us directly
   - name: Email
-    url: mailto:vuetorrent@larsluph.dev?cc=dw.daanwijns@gmail.com
+    url: mailto:vuetorrent@larsluph.dev
     about: Or sent us a mail for private requests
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -2,4 +2,7 @@ blank_issues_enabled: true
 contact_links:
   - name: Discord
     url: https://discord.gg/KDQP7fR467
-    about: 'Join the discord server to contact us directly'
+    about: Join the discord server to contact us directly
+  - name: Email
+    url: mailto:vuetorrent@larsluph.dev?cc=dw.daanwijns@gmail.com
```
WDaan, line 7: We probably should only rely on dedicated Vuetorrent email addresses and not personal ones.

WDaan (changes requested):

.github/ISSUE_TEMPLATE/config.yml
Larsluph: WDYM by dedicated email address? Do you have one we can use?

WDaan: Would prefer to not have a personal one in there.

Don't really have an alternative ready right now.

Larsluph: So I leave only mine?

WDaan: If you're okay with that, yes :)

WDaan (approved):

PR #1754, opened 2024-06-26
feat(PWA): Add *.torrent file handler

### LINKED ISSUE #989: [Feature Request] Add *.torrent file association for PWA
By adding this feature, This PWA will be a great remote-thin-client replacement. 
I'm guessing there might be similarities with the magnet link handler (#111 #784) in the approach of achieving this.

https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/handle-files

Larsluph:
Fixes #989

- Updated screenshots and included them in the manifest for the rich install UI: [source](https://developer.chrome.com/blog/richer-pwa-installation/)
- Added `file_handlers`, `launch_handler`, `protocol_handlers` and `shortcuts`
- Add launchQueue processing in the root app component to handle file associations

@stasink Do you have everything to test if it works?


Larsluph:
Ready to merge

@stasink I suppose your reaction means that it's working?

WDaan (approved):
Really cool to get this one in there 🚀 

Would the PWA need to be reinstalled for some of these manifest changes to make it through? 🤔

Larsluph:
> Would the PWA need to be reinstalled for some of these manifest changes to make it through? 🤔 

I have absolutely no idea as I don't use those but I suppose that browsers are automatically updating them on launch

Larsluph:
From [Stack overflow](https://stackoverflow.com/a/66783321), it seems that a manifest change will update the PWA automatically.

Larsluph:
No response, merging and we'll see...

PR #1756, opened 2024-06-27
feat(docker): Add official LSIO mod docker image to pipeline

Larsluph:
Create official linuxserver docker mod image

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -4,6 +4,10 @@ on:
     branches:
       - master
 
+env:
+  REGISTRY: ghcr.io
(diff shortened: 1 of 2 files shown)


WDaan:
What does a container with just  VueTorrent inside provide? 🤔

I was expecting a pre-mounted qbit one. What am I missing?

Larsluph:
That's basically what all other images provide, I suppose the linuxserver image will update to the latest update of the mod, at least on every restart.

I can see if providing a default qBittorrent configuration is possible. We just need to make sure that this doesn't overwrite existing configurations

WDaan (approved):

PR #1762, opened 2024-06-28
perf(network): Reduce API calls by using the maindata sync throughout the app

### LINKED ISSUE #1606: [Improvement]: Improve app performance by removing the extra info call
### Bug description

image
Please notice the requests marked with arrow, browser will keep fetching `torrents/info`. While I have a huge number of torrents, this behaviour will  occupy considerable resources of network and cpu/memory.

### Expectations

Stop calling this API periodly unless necessary，just like the default webui does.

_No response_

### Steps to reproduce

1. Open a fresh vuetorrent page
2. Observe the network panel of Devtools

Built from source

### Environment description

```plain text
Device type and name: PC
OS: Win11

PWA: No
Running inside a container: No
```

### qBittorrent version

4.6.2

### VueTorrent version

2.7.2

### Agreements

- [X] The bug still occurs on the nightly build
- [X] The bug still occurs on the latest version of qBittorrent
- [X] A similar issue doesn't already exist

Larsluph:
Rework app data flow to only use the maindata sync API call.
Fixes #1606

- Removed manual fetch when entering views, thus reducing load times.
- Fetch qbit version (and store it) instead of useless build info
- Split metadata store into several small ones to reduce circular imports
- Precise tracker filtering now that we get all torrents' tracker data


WDaan (approved):

WDaan:
Cool! We'll soak this a bit in the next release to see what weird things might pop up.

Good to have this redesign through, the first iteration was a bit naive and exploded over time with all these additional features.

flashlab:
@Larsluph  Glad to see this improvement happen :tada::tada: But the UI freeze and popup unresponsive error after page first load. I'm not sure if it related with the new sort system. And the seeds is over 5,200 in my qbittorrent(v6.7.1 Libtorrent v1.2.19).

Larsluph:
> UI freeze and popup unresponsive error after page first load. I'm not sure if it related with the new sort system.

Can you precise what is happening exactly? Do you mean that the entire app is slowed down?

The new sort system shouldn't be a problem, only view render is. During testing I noticed a lot of slowdowns but I think those were due to mocks changing all data, thus reloading the whole view each time.

I measured about 300ms for raw sorting and 5-9 secs for rendering with 15000 torrents.

flashlab:
@Larsluph After some tests, I found the issue only occurs under seeding state. If all seeds are paused, the UI will not be stucked. And here is the network panel during the freeze:
image
The routine api query delayed a lot. I can see the list of torrents, but totally no response whether scroll the page or click on anywhere.

Larsluph:
> issue only occurs under seeding state. If all seeds are paused, the UI will not be stucked.

That seems logical given that if nothing changes, qBittorrent will send an empty response, no effect will be applied to the app so no freeze.

> The routine api query delayed a lot. I can see the list of torrents, but totally no response whether scroll the page or click on anywhere.

I think a new wait time starts after the end of the last one. If the view is freezing, this means that a new wait will be initiated when the app goes back from freeze.

I need to check whether keeping the view still while updating data in the background still introduces those freeze. If that's the case we'll need some improvement to the tracking performed by Vue.

PR #1771, opened 2024-06-30
perf(content): Group numbers when sorting filenames

Larsluph:
Improve sort in tree builder to use group of numbers instead of raw string sort.

## Example

| Index | Before   | After    |
|-------|----------|----------|
| 1     | [aaabbb] | 1_aaa    |
| 2     | 1_aaa    | 2_aaa    |
| 3     | 10_aaa   | 10_aaa   |
| 4     | 11_aaa   | 11_aaa   |
| 5     | 2_aaa    | [aaabbb] |
| 6     | aaabbb   | aaabbb   |


Larsluph:
Just a pending thought before merging, should we use that new sort system for all text values, or should we keep two different?

Larsluph:
I have literally no idea why `this` were resolved to `undefined`, something about class scopes maybe???
I anyone has an explanation, I'll take it 😂

WDaan (approved):
I would say keep them different as they serve a different purpose? 🤔

Issue #1173, opened 2023-10-19
[Feature Request] Torrent Content: Filter files by Size


tainewoo:
[Feature Request] Torrent Content: Filter files by Name/Size
Currently all the files are all listed there.
If we can Filter them by name and size, that would be more easier for users to organize the downloading/uploading (together with another needed feature: bulk edit to check/uncheck the files)

tainewoo:
filter files by name is implemented， it would be awesome if we can filter files by size 😀 
![image](https://github.com/VueTorrent/VueTorrent/assets/13391232/6012cb9a-82d9-47d5-8cb9-9f6a58944e43)

tainewoo:
Hi @Larsluph 
Will u consider this feature 😀

Larsluph:
> Hi @Larsluph 
> Will u consider this feature 😀

The design still has to be reworked, I was thinking about creating a dialog to manage the currently applied filter but I'm still not sure how to display the activator, especially with the resume/pause timer button.

Larsluph:
This would work for sorting but not filtering.

Maybe a FAB can be a great idea to allow for both without taking too much space.

WDaan:
Fab was the word I was looking for! 😄 

I guess sorting would already help out a bit? Not sure how you would even use filtering on size? 🤔 
Would a user pick a filter between size ranges? What would the granularity look like?

Larsluph:
@tainewoo I'm currently working on it with everything I have in mind. I was wondering if you think the filter is still necessary if we introduce bulk selection? It will add a bit of complexity so I need to think about it if we want to keep it.

Here's a little sneak peak 👀 
![image](https://github.com/VueTorrent/VueTorrent/assets/22910497/53721c98-559f-442f-bb79-967c43b54b4e)

Larsluph:
Bulk selection of files using criterias including size has been added, not sure if filters are still wanted or if bulk is enough.

tainewoo:
Hi @Larsluph ,

Sorry for the late reply, I was away from keybord last month (while my qB is running automatically relied on the great features of VueTorrent webUI.)
Thats cool, it's pretty good for me to filter files if the torrent has houndred of files in it.
Usually I will need to partial seeding some torrents to speed up them, because my harddisk is not big enough but I want to help seed more torrents.

Either sorting or filter  or this bulk selection filter can help on this use scenario. Now we have this bulk selection filter, I can consider this feature is completed, thank you all so much!

[...]

PR #1795, opened 2024-07-19
perf(content): Add bulk selection using filters

Larsluph:
#1173

This PR creates a new dialog that allows the user to bulk select files for a torrent.

image


WDaan (approved):
Awesome 🙌

PR #1796, opened 2024-07-19
perf: Remove openSideBarOnStart setting in favor of screen type

Larsluph:


WDaan:
Some people wanted this though 🤔 IIRC

Larsluph:
In that case we need to make a difference for screen type configuration.
I'm currently used to having the navbar opened on desktop and closed on mobile.

I'm not sure about the real added value of opening the navbar automatically on start on mobile...
Can you find the original issue if any?

WDaan:
I think it was the other way around, people that want it closed on desktop.

But not sure if important enough to care

WDaan (approved):

PR #1797, opened 2024-07-20
perf: Drastically increase performance on dashboard and content view

### LINKED ISSUE #1786: [Perf] UI freezed while seeding 5000+ torrents
As first descirbed in #1762 , if more than 5000 torrents are under seeding state, the web UI will freeze right after torrent list show up. And all mouse/clipboard interactions are non-responsible.
If all seeds are paused, the issue will gone.
After applying the latest patch #1784 , the issue still exists.

I suppose if the regular UI  updating job(such as sorting) can be disabled if seeding counts exceed a certain number(let's say 1000), and only updating if filter state changes which triggered manually ;)

Larsluph:
Fixes #1786

### DIFF
--- a/src/components/TorrentDetail/Content/Content.vue
+++ b/src/components/TorrentDetail/Content/Content.vue
@@ -11,7 +11,7 @@ const props = defineProps<{ torrent: Torrent; isActive: boolean }>()
 
 const { height: deviceHeight } = useDisplay()
 const contentStore = useContentStore()
-const { rightClickProperties, filenameFilter, openedItems, flatTree, internalSelection, timerForcedPause, isTimerActive } = storeToRefs(contentStore)
+const { rightClickProperties, filenameFilter, flatTree, internalSelection, timerForcedPause, isTimerActive } = storeToRefs(contentStore)
 const dialogStore = useDialogStore()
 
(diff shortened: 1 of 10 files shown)


WDaan:
Why the openedItems change?

Surprised to see ShallowRef making that big of a difference, but small things add up with the amount of torrents

Good find 🚀

WDaan (approved):

Larsluph:
> Why the openedItems change?

At first it was only to use shallowRef on all reactive objects but that might add overhead for nothing. We can revert that.

Larsluph:
> Why the openedItems change?

Looking back at it to fix and I think what you meant was for props change to store ref?

If that's the case, Vue disallows not mutating props values by design:

> This means you should not attempt to mutate a prop inside a child component.

https://vuejs.org/guide/components/props.html#one-way-data-flow

As the data is available inside a store I just moved it to the child component.

alzheimerspenguin:
This PR made VueTorrent usable for me, thank you so much.
The difference is massive.

Larsluph:
> This PR made VueTorrent usable for me, thank you so much. The difference is massive.

Huge thanks, this means a lot 🙏 
Glad you like it 😄

PR #1800, opened 2024-07-22
perf: Display speed limits in navbar

### LINKED ISSUE #1726: [Feature Request]: Show globale rate/speed speed limit in speed cards and graph
### Description

Hi,
I just switched to VueTorrent and love it so far. One thing I'm missing though is to show the current global rate/speed limit somewhere as I'm used to have it shown on the classic Web UI as I have some scripts controlling the download and upload speed limit based on different metrics. Would be nice to have these limits shown on the Dashboard speed cards and graph to be sure the limits are automatically set correct

Having the limits in parantheses in the speed cards and colored horizontal lines for the limits in the speed graph

_No response_

Larsluph:
Fixes #1726

@Seros
Here's my shot at adding the speed limits to the UI, lmk what you think.

image

### DIFF
--- a/src/components/Navbar/SideWidgets/SpeedGraph.vue
+++ b/src/components/Navbar/SideWidgets/SpeedGraph.vue
@@ -25,7 +25,7 @@ const chartOptions: ApexOptions = {
       enabled: false
     }
   },
-  colors: ['#f38a09', '#f38a09', theme.current.value.colors.upload, theme.current.value.colors.download],
+  colors: [theme.current.value.colors.upload, theme.current.value.colors.download, theme.current.value.colors.upload, theme.current.value.colors.download],
   stroke: {
     show: true,
     curve: 'smooth',
--- a/src/components/Navbar/SideWidgets/SpeedGraph.vue
+++ b/src/components/Navbar/SideWidgets/SpeedGraph.vue
@@ -1,8 +1,9 @@
 <script setup lang="ts">
 import { formatSpeed } from '@/helpers'
-import { useNavbarStore, useVueTorrentStore } from '@/stores'
-import { ApexOptions } from 'apexcharts'
 import dayjs from '@/plugins/dayjs'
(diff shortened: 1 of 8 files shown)


Seros:
That looks pretty fine to me, thanks a lot! I'm not sure if different colors for the limits would make sense though but that's no big deal

src/components/Navbar/SideWidgets/SpeedGraph.vue
```diff
@@ -23,21 +25,22 @@ const chartOptions: ApexOptions = {
...
     }
   },
-  colors: [theme.current.value.colors.upload, theme.current.value.colors.download],
+  colors: ['#f38a09', '#f38a09', theme.current.value.colors.upload, theme.current.value.colors.download],
   stroke: {
     show: true,
     curve: 'smooth',
     lineCap: 'round',
-    width: 4
+    width: 3,
```
Seros, line 33: I tried to work on this myself and had some issues applying different widths on the lines in the graph.  Every time I did the whole graph just disappeared. I suppose this is a general issue coming from the apexcharts libary or?

WDaan:
This is a really cool addition! 😄 

I would perhaps make this one optional though? I assume some people would like their sidebar to stay lean and simple.

src/components/Navbar/SideWidgets/SpeedGraph.vue
Larsluph: Yeah I had the same issue, some error in the SVG path pops up when using an array, even though the parsing is done correctly...

Larsluph:
> I would perhaps make this one optional though?

Only the graph series or the additional cards as well? For now the limit cards are only displayed if a limit is set.

Larsluph:
> I'm not sure if different colors for the limits would make sense

I already tried before but didn't think about dashed lines so it was confusing, with dashes it's really good.

image

Larsluph:
Also added partial display of limit cards, not sure if it's relevant...

image

WDaan (approved):

PR #1826, opened 2024-08-03
perf: Add configurable format for duration values

### LINKED ISSUE #1818: [Feature Request]: Seeding time editable date format
### Description

Currently the qBittorrent details for the seeding time show it in the table view as a duration in the format of e.g. 2M 1d 5h 10m 30s. Considering this field is informative, but the level of it is too detailed, it'd be great if we had a duration formatter in the settings.

I can only presume, but being a duration field, it's probably very easy to have this formatted from a pattern. 

Maybe it doesn't have to be a formatting pattern, but pre-written formats like "MM dd hh" or the full one.

Larsluph:
Fixes #1818

Use configurable format for ItemDuration (in Dashboard).

PanelDuration uses the humanize method from dayjs, should we keep it or also move to the same format?

### DIFF
--- a/src/components/Dashboard/DashboardItems/ItemDuration.vue
+++ b/src/components/Dashboard/DashboardItems/ItemDuration.vue
@@ -1,29 +1,21 @@
 <script setup lang="ts">
-import dayjs from '@/plugins/dayjs'
+import { formatDuration } from '@/helpers'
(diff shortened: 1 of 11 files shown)


WDaan (approved):
Nice!

I think the more things we can tackle the same way, the better.

WDaan (approved):
Cool 😎

PR #1834, opened 2024-08-04
feat: Add global error handler

Larsluph:


Larsluph:
> I think this would be a good idea.

I just need to test if all errors are logged or if only start up errors are shown as this may be too verbose on slow/unreliable networks but hte idea's there!

WDaan (approved):

PR #1870, opened 2024-08-22
feat: Show or hide top or bottom pagination bars

martylukyy:
# Feature: Show or hide top or bottom pagination bars

This PR adds two checkboxes to the VueTorrent General settings that allow the user to only how the top or the bottom pagination bar.
Once one bar gets hidden, by unticking the checkbox, the other checkbox will get disabled so that the user cannot disable both pagination bars, which could lead to an upsurge in support requests. 

## PR Checklist

- [x] All tests pass


src/components/Dashboard/Toolbar.vue
```diff
@@ -152,9 +153,10 @@ function toggleSelectMode() {
     </div>
 
     <v-spacer v-if="$vuetify.display.mobile" />
-    <v-pagination v-else class="flex-grow-1 overflow-hidden" v-model="currentPage" :length="pageCount" next-icon="mdi-menu-right" prev-icon="mdi-menu-left" />
+    <v-pagination v-if="showTopPagination && !$vuetify.display.mobile" class="flex-grow-1 overflow-hidden" v-model="currentPage" :length="pageCount" next-icon="mdi-menu-right" prev-icon="mdi-menu-left" />
 
-    <div class="d-flex align-center text-uppercase text-select" style="font-size: 0.8em">
+    <div class="flex-grow-1" v-if="!showTopPagination"></div>
```
Larsluph, line 158:
```vue
<v-pagination v-if="showTopPagination && !$vuetify.display.mobile && pageCount > 1" class="flex-grow-1 overflow-hidden" v-model="currentPage" :length="pageCount" next-icon="mdi-menu-right" prev-icon="mdi-menu-left" />
<v-spacer v-else />
```
You can reduce the number of condition branches and elements.

martylukyy: Thanks this is the first time touching something Vue for me.
So the v-spacer/if/else part doesn't come naturally to me. 😂

src/components/Settings/VueTorrent/General.vue
```diff
@@ -163,6 +163,13 @@ function openBackendHelp() {
           <v-checkbox v-model="vueTorrentStore.isDrawerRight" hide-details density="compact" :label="t('settings.vuetorrent.general.isDrawerRight')" />
         </v-col>
 
+        <v-col cols="12" sm="6">
+          <v-checkbox v-model="vueTorrentStore.showTopPagination" :disabled="!vueTorrentStore.showBottomPagination" hide-details density="compact" :label="t('settings.vuetorrent.general.showTopPagination')" />
+        </v-col>
+        <v-col cols="12" sm="6">
+          <v-checkbox v-model="vueTorrentStore.showBottomPagination" :disabled="!vueTorrentStore.showTopPagination" hide-details density="compact" :label="t('settings.vuetorrent.general.showBottomPagination')" />
+        </v-col>
```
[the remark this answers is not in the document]
martylukyy: Done.
I tried to create an solution that is as concise as possible.
If you want the implementation more readable with more lines of code, please tell me.

martylukyy:
> Thanks for the contribution! However I'm not sure that we're gonna accept it.
> We just removed the ability to switch between top and bottom pagination to always make it both. Are you aware of that change? Is this PR related?

No this PR is not related to it.
I just recently switched to VueTorrent and the infinite scroll function won't work for me due to the amount of elements in my client i think.. 🤔  VueTorrent just crashes upon loading the table view with all the elements...
So the pagination option is the only option for me. 😅 
Ideally i want my torrent table not to overflow at all so i don't need to scroll. (with pagination active) 
That is why the bottom pagination is a nuisance to me here.

Larsluph (approved):
I just reworded the settings' select values and added some icons for less text and more visuals!

Thanks again for the contribution!

Larsluph:
Also changed from 2 booleans to a single flag to prevent "complicated" setters and provide completion

Issue #1840, opened 2024-08-08
[Feature Request]: Add a panel with torrent informations


pokemaster974:
### Description

Hi, I've searched into the VueTorrent's settings but haven't found such setting.
Is it possible to add an option that when we click on a torrent it will display directly an informations panel (like in the default panel) ?
The panel could be se same as with right click -> Show info or a more condensed panel.

An alternative could be that a double click will display this panel (actually it seems that double click made a select/unselect).
Regards.

### Complementary informations

- [X] Is this feature already implemented in the default WebUI?

pokemaster974:
Ok, I understand that I need to uncheck "select mode" to have the double click working.
I just let it open to have your feedback about a below permanent panel showing the torrent informations. And feel free to close it if necessary.

Larsluph:
Do you mean something like in #1638? If not, I'm not sure to understand your request.

pokemaster974:
> Do you mean something like in #1638? If not, I'm not sure to understand your request.

I mean something like the original UI :
![image](https://github.com/user-attachments/assets/f38fbde1-6637-4e61-8017-6a13c950a941)

I'm thinking of something like that :
![image](https://github.com/user-attachments/assets/f35cb09d-1ea9-4629-ad0c-a19a002c1f4a)
And when we select a torrent, it refreshes and displays the informations about the selected torrent.

pokemaster974:
The issue I see with the current behaviour is that informations panel is displayed on all the webUI and we need an extra click (after the double click or the right-click -> show info) to close it. Having a bottom panel removes this extra step to close it.

Larsluph:
Oh ok now I understand. I like the idea of removing that click barrier but I don't know if it's really something we should do.

For the table view on desktop it does looks great, but for all the other configurations, whether on mobile or with the list and grid view it doesn't look great as it takes a lot of space.

PR #1880, opened 2024-09-01
feat: Highlighting and peeking torrent

huyld:
# Highlight torrent and display details
In this PR I tried to mimic the info panel of the original WebUI. Currently, left clicking on torrent does nothing, just quickly flashes the clicked torrent. So I want to try out an idea of peeking torrent details without having to navigated to another page. In other to do that I introduced these changes:

## Highlighting torrent
In Dashboard view, user can highlight a single torrent by left clicking on it. This is done in when the select mode is OFF.
The torrent hash is kept in the `dashboard` store's `highlightedTorrent`. In the dashboard, it would have the color corresponded to its state, with the variant "-lighten-3". Torrents can be highlighted in all views (Table, List & Grid).

## Refactoring of `TorrentDetail.vue`
`TorrentDetail` now has the optional prop `isPeeking` (false by default).

Normally, double clicking on torrent will redirect user to the detail route and display `TorrentDetail` as a full page. The torrent hash and the active tab is retrieved from the route. Changing tabs modified the route as usual.

Now when torrent is highlighted, Dashboard.vue renders `TorrentDetail` inside a bottom sheet, passing `isPeeking` true to this component. The torrent hash is retrieved from `dashboard` store's `highlightedTorrent`. TorrentDetail keeps the active tab internally. Changing tabs won't change route.

Clicking outside of the bottom sheet simply closes it.

### Some screenshots
peeking01
peeking02
peeking03

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Dashboard/Views/Grid/GridTorrent.vue
+++ b/src/components/Dashboard/Views/Grid/GridTorrent.vue
@@ -71,12 +71,7 @@ function getTorrentColor(torrent: Torrent) {
 </script>
 
 <template>
-  <v-card
-    class="cursor-pointer"
-    :style="`border-left: 6px solid ${stateColor}`"
-    height="100%"
-    :color="getTorrentColor(torrent)"
-    @click="$emit('onTorrentClick', $event, torrent)">
+  <v-card class="cursor-pointer" :style="`border-left: 6px solid ${stateColor}`" height="100%" :color="getTorrentColor(torrent)" @click="$emit('onTorrentClick', $event, torrent)">
     <v-card-title class="text-wrap text-subtitle-1 pt-1 pb-0">{{ torrent.name }}</v-card-title>
     <v-card-text>
       <div class="d-flex flex-gap flex-wrap">
--- a/src/components/Dashboard/Views/Grid/GridTorrent.vue
+++ b/src/components/Dashboard/Views/Grid/GridTorrent.vue
@@ -59,6 +59,15 @@ const getComponent = (type: DashboardPropertyType) => {
 
 const isTorrentSelected = computed(() => dashboardStore.isTorrentInSelection(props.torrent.hash))
 const stateColor = computed(() => current.value.colors[getTorrentStateColor(props.torrent.state)])
+
+function getTorrentColor(torrent: Torrent) {
+  if (isTorrentSelected.value) return `${getTorrentStateColor(torrent.state)}-darken-3`
+  if (dashboardStore.isSelectionMultiple) return undefined
+  if (dashboardStore.isTorrentHighlighted(torrent.hash)) {
+    return `${getTorrentStateColor(torrent.state)}-lighten-3`
+  }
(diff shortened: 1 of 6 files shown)


Larsluph:
#1840

Larsluph:
The only issue I found by fidling around is the `files` request failing because of the lacking hash in the URL.
Request triggered [here](https://github.com/huyld/VueTorrent/blob/feat/highlighting-and-peeking-torrent/src/pages/TorrentDetail.vue#L95).

Apart from that everything looks good from a functional perspective.
I'll take a deeper look into the PR when I have time.

Larsluph:
I took a look at the changes you introduced and I have to say that I have mixed feelings about these changes.

Looking back at what was originally asked in the tagged issue, the user wanted to remove the extra click needed to close the TorrentDetail view, which I totally understand when you need to gather some quick info about your torrents.
This PR introduces both the same extra click AND a timeout which may be worse. On top of that, it requires some changes in the existing codebase that creates new dependencies between components.

Without a simpler interface to peek at, I don't really see the added value that would compensate for all the changes it requires.

huyld:
Thanks for the feedback. It's exactly what I want to get for this PR. Sorry for the delayed response.

As stated in the description I wanted to try another idea to look at torrent detaisl without navigating to a new page. I should have clarified more that my personal problem with it is not the click count, but it makes me lose track of the torrent when I go back to the table view (I have many torrents with large pagination size).

I'm aware that a bottom info panel would take up many space. Hence, I use the bottom sheet so that it wouldn't affect the general layout.
Without a dedicated area for displaying torrent info, I think it's pretty hard to avoid that extra click. I'd also love to hear any idea that help us circumvent this problem.
Perhaps we can try having such static area only for large screens. Mobile and smaller screen users don't really need to see this. Wdyt?

And honestly I don't like the way `TorrentDetail.vue` being refactored in the PR neither. I feel like I'm "hijacking" this page to ignore the route and behave as a component with internal tab state. I just wanted to bring the change to get some feedback before making more changes to the code base. I do agree `TorrentDetail.vue` needs another iteration before it could be merged.

Larsluph:
@huyld I'm also busy and can't really focus on VueTorrent ATM so no worries!

I was missing your point and was thinking more about the linked issue, so the extra click is good.

If you just want to keep the scroll in place while changing pages, maybe we don't need to refactor anything at all!

I took a look at `vue-router`'s documentation and it looks like we can control scroll behavior between pages however we want (which isn't by default):
https://router.vuejs.org/guide/advanced/scroll-behavior.html#Scroll-Behavior

WDaan:
What will happen with this one??

Larsluph:
After some tests, the scroll behavior's savedPosition only applies when using the browser back action (so not with the X button at the top of the view). We'll have to make our own handler for this to work regardless of this state.

Anyway this would introduce too much change in the current codebase so this PR won't be merged in this state.

PR #1886, opened 2024-09-04
perf: Add catch-all route

Larsluph:
Prevent empty page if URL doesn't exists. Redirects to the dashboard page.


WDaan (approved):
Lol surprised we didn't have that already

PR #1928, opened 2024-09-30
feat: Support qBittorrent v5

Larsluph:
Now that qBittorrent v5 is officially released, we can start implementing new features. Let's start by fixing breaking changes.

Using `git diff release-4.6.7...release-5.0.0 -- src\webui\api` to generate changes in WebAPI.

- [X] Update types and deprecate old symbols
- [X] Fix breaking changes (start/stop)
- [X] Implement search plugin download
- [X] Display new search plugin result properties
- [X] Implement test email
- [ ] Add new settings
- [ ] Support getDirectoryContent, use it instead of local history if on qbit 5.0


Larsluph:
Build is not working with new types, v5 features will be implemented in another PR to prevent further reports from users.

WDaan:
Is this PR breaking? It looks like it isn't but I'm afraid about V4 setups breaking.

Larsluph:
I tried keeping everything by only adding new stuff and deprecating others. I'm in the process of reviewing everything to make sure I did keep the old compatibility.

I was also thinking about making a baseline for qBittorrent 4 users that may be unable to update or that just don't want to. So that we can focus on qBit 5 without thinking about retro-compatibility.

jaconey:
Thanks for fixing. Any chance to get a nightly build for this? Currently the main branch is broken for me on 5.0. If we are merging it soon, I'll just wait and use the vanilla UI to unblock. thx

WDaan (approved):
Apologies for the delay, looks fine 👌

Let's go v5 🚀

PR #1934, opened 2024-10-01
chore(CI): Use custom action for Discord release alerts

Larsluph:


WDaan (approved):

WDaan:
Why this change?? 🤔

Larsluph:
> Why this change?? 🤔

The action we currently use formats the release body on links, however Discord now natively supports links.

Both the release title link and the formatter they use clashes each other and creates more []/() couples which breaks the Markdown renderer.

PR #1958, opened 2024-10-08
  [fix] The peer info misplaced when no country flag returned.

### LINKED ISSUE #1872: [Bug]: A Peer IP without a Country Flag Messes up Peers View
### Bug description

Inside the Peers tab of a torrent's information, an IP that can't be resolved with a country is displayed incorrectly in the table of peers. Please see the image below.

bug-report

### Expectations

I expected to see either an N/A or a blank value for the peer and all of the other information to be displayed in the correct rows, like the rows below.

TypeError: undefined has no properties ["item.country"]

The above error was displayed in every update frame.

### Steps to reproduce

1. Manually add a peer with a local IP (192.xxx:xxxx) to a torrent.
2. Continue to view the peer tab of torrent info while it gets the peer's country flag.
3. All values in the row are offset by one from the respective columns, see the above image.

Using a third-party docker image (hotio or others)

### Environment description

```plain text
Device type: Desktop
OS: Windows
Browser (name and version): Firefox 129.0.2 (64-bit)

PWA: Yes
Running inside a container: Yes
```

### qBittorrent version

4.6.6

### VueTorrent version

2.12.0

### Agreements

- [X] The bug still occurs on the nightly build
- [X] The bug still occurs on the latest version of qBittorrent
- [X] A similar issue doesn't already exist
- [X] I checked the DevTools and included any errors that may show up

AD590:
Fixes #1872

Occasionally the peer host gets the gateway IP which is a local IP address(as 192.168.0.1). Check whether to add blank before getting the flag.

Screenshot 2024-10-08 134725

### DIFF
--- a/src/components/TorrentDetail/Peers.vue
+++ b/src/components/TorrentDetail/Peers.vue
@@ -162,13 +162,9 @@ onBeforeRouteUpdate(() => !addPeersDialog.value)
 
       <template #[`item.country`]="{ item }">
         <div v-if="showCountryFlags" class="cursor-help" :title="item.country">
-          <template v-if="item.country_code">
-            <img v-if="isWindows" :alt="codeToFlag(item.country_code!).char" :src="codeToFlag(item.country_code!).url" :title="item.country" style="max-width: 32px" />
-            <span v-else :title="item.country">{{ codeToFlag(item.country_code!).char }}</span>
-          </template>
-          <template v-else>
-            <span></span>
-          </template>
+          <span v-if="!item.country_code"></span>
+          <img v-else-if="isWindows" :alt="codeToFlag(item.country_code).char" :src="codeToFlag(item.country_code).url" :title="item.country" style="max-width: 32px" />
+          <span v-else :title="item.country">{{ codeToFlag(item.country_code).char }}</span>
         </div>
       </template>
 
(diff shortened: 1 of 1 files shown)


src/components/TorrentDetail/Peers.vue
```diff
@@ -162,8 +162,13 @@ onBeforeRouteUpdate(() => !addPeersDialog.value)
...
         <div v-if="showCountryFlags" class="cursor-help" :title="item.country">
-          <img v-if="isWindows" :alt="codeToFlag(item.country_code!).char" :src="codeToFlag(item.country_code!).url" :title="item.country" style="max-width: 32px" />
-          <span v-else :title="item.country">{{ codeToFlag(item.country_code!).char }}</span>
+          <template v-if="item.country_code">
+            <img v-if="isWindows" :alt="codeToFlag(item.country_code!).char" :src="codeToFlag(item.country_code!).url" :title="item.country" style="max-width: 32px" />
+            <span v-else :title="item.country">{{ codeToFlag(item.country_code!).char }}</span>
+          </template>
+          <template v-else>
+            <span></span>
+          </template>
```
Larsluph, line 171: No need for an extra template, just slap the condition on top of others. With this, we can remove the non-null assertion operator.

```suggestion
          <span v-if="!item.country_code"></span>
          <img v-else-if="isWindows" :alt="codeToFlag(item.country_code).char" :src="codeToFlag(item.country_code).url" :title="item.country" style="max-width: 32px" />
          <span v-else :title="item.country">{{ codeToFlag(item.country_code).char }}</span>
```

Larsluph:
Fixes #1872

Larsluph (approved):

PR #1998, opened 2024-10-28
fix(ActiveFilters): Force-enable filter after clear

### LINKED ISSUE #1995: [Feature Request]: Removing one or more filters filters should reset the category and/or the global state
### Description

When a filter active, it can be removed by pressing the 'remove filter' button. But if the filter had been disabled before it was removed, then when the same filter or a different one of the same type are (re)added they will start off as disabled, and the user will have to manually toggle them as enabled by clicking the 'toggle' button.

If all the active filters are disabled before they are removed, then *any* new filter that is added will start off as disabled.

Example 1: if I filter the torrents by "category 1", disable that filter, remove it, and then try to filter by "category 2", the new filter will be added as disabled because when I disabled "category 1" the disabled state was applied to the whole "Category" filter type (which is a whole other can of worms) and, more importantly, that disabled state was not cleared when I removed the "category 1" filter.

Example 2: If I filter my torrents with multiple filters, disable all of them globally, remove them, and then add one or more new filter, the new filters will start off as disabled because the global filter list is still set as disabled.

When filters are removed one by one, their respective category/type should be unconditionally reset 'enabled'.

When all filters are removed at once, the global filter state should be unconditionally reset to 'enabled'.

_No response_

Larsluph:
Fixes #1995

### DIFF
--- a/src/stores/torrents.ts
+++ b/src/stores/torrents.ts
@@ -6,7 +6,7 @@ import { useTorrentDetailStore } from '@/stores/torrentDetail.ts'
 import { RawQbitTorrent } from '@/types/qbit/models'
 import { AddTorrentPayload } from '@/types/qbit/payloads'
 import { Torrent as VtTorrent } from '@/types/vuetorrent'
-import { useArrayFilter, useSorted } from '@vueuse/core'
+import { useArrayFilter, useSorted, whenever } from '@vueuse/core'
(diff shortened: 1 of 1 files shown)


WDaan (approved):

Larsluph:
> Now all disabled filters get re-enabled on launch. 

@romartobashey
Please open an issue with valid reproduction steps

PR #2021, opened 2024-11-04
feat: Adding secondary sorts to peers and seeds

### LINKED ISSUE #1889: [Feature Request]: Seeds & Peers column sorting - use total value as secondary sort
### Description

Sorting by the `Seeds` column only sorts by currently connected seeds. This makes it redundant when no torrents are currently seeding to a leecher.

VueTorrent (seemingly random order):
msedge_2024-09-09_00-34-32

The same issue applies to the `Peers` column.

In the default qBittorrent WebUI & ruTorrent, sorting by the `Seeds` column sorts first by currently connected seeds, but then also by total seeds.

qBittorrent default WebUI:
msedge_2024-09-09_00-39-54

ruTorrent:
msedge_2024-09-09_00-33-58

Note that I didn't have any currently connected seeds in these screenshots, but if I did, they would have been sorted to the top first, regardless of total seed number.

_No response_

JWWolstenholme:
# Adding secondary sorts to peers and seeds

Closes #1889.

This commit modifies the sorting logic for the `num_leechs` & `num_seeds` torrent sort keys.

The original sorting logic is maintained, however when torrents are "tied", the `available_peers` & `available_seeds` sort keys are used to further sort.

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/constants/vuetorrent/Comparators.ts
+++ b/src/constants/vuetorrent/Comparators.ts
@@ -33,16 +33,8 @@ const comparatorMap: Record<keyof Torrent, (a: Torrent, b: Torrent, isAsc: boole
   last_activity: (a, b, isAsc) => comparators.invertedNumeric.compare(a.last_activity, b.last_activity, isAsc),
   magnet: (a, b, isAsc) => comparators.text.compare(a.magnet, b.magnet, isAsc),
   name: (a, b, isAsc) => comparators.textWithNumbers.compare(a.name, b.name, isAsc),
-  num_leechs: (a, b, isAsc) => {
-    const primaryComparison = comparators.numeric.compare(a.num_leechs, b.num_leechs, isAsc)
-    if (primaryComparison === 0) return comparatorMap['available_peers'](a, b, isAsc)
-    else return primaryComparison
-  },
-  num_seeds: (a, b, isAsc) => {
-    const primaryComparison = comparators.numeric.compare(a.num_seeds, b.num_seeds, isAsc)
-    if (primaryComparison === 0) return comparatorMap['available_seeds'](a, b, isAsc)
-    else return primaryComparison
-  },
+  num_leechs: (a, b, isAsc) => comparators.numeric.compare(a.num_leechs, b.num_leechs, isAsc) || comparators.numeric.compare(a.available_peers, b.available_peers, isAsc),
+  num_seeds: (a, b, isAsc) => comparators.numeric.compare(a.num_seeds, b.num_seeds, isAsc) || comparators.numeric.compare(a.available_seeds, b.available_seeds, isAsc),
   priority: (a, b, isAsc) => {
     if (a.priority > 0 && b.priority > 0) return comparators.numeric.compare(a.priority, b.priority, isAsc)
     else if (a.priority <= 0 && b.priority <= 0) return comparators.numeric.compare(a.added_on, b.added_on, isAsc)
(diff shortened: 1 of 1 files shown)


JWWolstenholme:
Before this PR:
image

After this PR:
msedge_2024-11-05_04-11-16

src/constants/vuetorrent/Comparators.ts
```diff
@@ -33,8 +33,16 @@ const comparatorMap: Record<keyof Torrent, (a: Torrent, b: Torrent, isAsc: boole
...
+  num_leechs: (a, b, isAsc) => {
+    const primaryComparison = comparators.numeric.compare(a.num_leechs, b.num_leechs, isAsc)
+    if (primaryComparison === 0) return comparatorMap['available_peers'](a, b, isAsc)
+    else return primaryComparison
+  },
+  num_seeds: (a, b, isAsc) => {
+    const primaryComparison = comparators.numeric.compare(a.num_seeds, b.num_seeds, isAsc)
+    if (primaryComparison === 0) return comparatorMap['available_seeds'](a, b, isAsc)
+    else return primaryComparison
+  },
```
Larsluph, line 45: Awesome!
I personally don't like the reference to itself, even if it requires duplication.
You can also chain fallback comparisons using the OR (`||`) operator 👀

WDaan (approved):

src/constants/vuetorrent/Comparators.ts
```diff
@@ -33,8 +33,16 @@ const comparatorMap: Record<keyof Torrent, (a: Torrent, b: Torrent, isAsc: boole
...
-  num_seeds: (a, b, isAsc) => comparators.numeric.compare(a.num_seeds, b.num_seeds, isAsc),
+  num_leechs: (a, b, isAsc) => {
+    const primaryComparison = comparators.numeric.compare(a.num_leechs, b.num_leechs, isAsc)
+    if (primaryComparison === 0) return comparatorMap['available_peers'](a, b, isAsc)
+    else return primaryComparison
+  },
+  num_seeds: (a, b, isAsc) => {
+    const primaryComparison = comparators.numeric.compare(a.num_seeds, b.num_seeds, isAsc)
+    if (primaryComparison === 0) return comparatorMap['available_seeds'](a, b, isAsc)
+    else return primaryComparison
```
WDaan, line 44: Useless else?

WDaan (changes requested):

src/constants/vuetorrent/Comparators.ts
JWWolstenholme: Yup. I was just following the style of the sort key after this one which also has a redundant else:
https://github.com/VueTorrent/VueTorrent/pull/2021/files#diff-2731caa080eabd74ef3cbcccb750ede8eba9d53e9439c5e915ab13e8e55ce446R50

If I don't hear back from you I'll remove it as suggested when I come back to this PR to make these changes.

src/constants/vuetorrent/Comparators.ts
JWWolstenholme: Fair enough. I'll opt for duplication instead as suggested. And thanks for the tip about chaining, I'll use that too when I come back to this PR to make these changes.

src/constants/vuetorrent/Comparators.ts
Larsluph: For conditional return with only 2 conditions, I would personally use a terminary operator instead but an if-else is still acceptable.

The case you mentioned is a bit particular because it's a 4-case condition.

Anyway, it gets compiled to ternaries anyway by vite.

Larsluph (approved):
Thanks for the contribution!

WDaan (approved):

Issue #1401, opened 2023-12-09
[Feature Request] Ability to resize columns in table view.


SawkeeReemo:
Short and sweet: I love the rebuild, but would it be possible to be able to resize the column in table view at some point? Would be a welcome addition. Thank you for considering!

tainewoo:
Or maybe to set the width of the "torrent title" column, by percentage?
Currently the torrent title column takes too much width percentage out of the total browser width.

If it's set in percentage, it can also fits different window scales if users switch window mode and maxmized mode and full screen mode for the browser.

iamse7en:
+1 for this feature! Titles take way too much real estate for me in both desktop and mobile settings. One long title can widen the column far beyond what should be. I like how long titles skip to next line, but should do so much earlier in character count. Thanks for all the work on this theme, it's the very best!

i0Z3R0:
Would love this feature. Titles take up way too much space with all the info in the title. A simple drag resize bar between the labels (title, size, ratio, etc.) would be great.

puddingsoda:
+1 for me here, most of the time I want to quickly hide torrent names, and in qBittorrent it's very easy to change the title tab to 0% and mouse over if I want to know what a particular torrent is.

theolaa:
I agree as well, the title takes up way too much width of each row (probably 40% just for title on my 1080p monitor), and honestly I usually only care about the first few dozen characters or so (40 maybe)? Resizable columns would be very welcome!

swwrww:
Hi, just wanted to chime in and strongly support this request.

Being able to manually resize columns in Table view would significantly improve usability, especially on different screen sizes or when prioritizing specific fields like Name, Status, or ETA. Fixed-width columns can be quite limiting in everyday use.

It would also be great if column widths could be persisted across sessions (e.g. saved in localStorage or config), so users don’t have to keep adjusting them.

This enhancement would bring VueTorrent closer to the flexibility of traditional desktop clients, while keeping the clean UI.

Thanks for considering! 🙏

kevingandrades:
+1 for this feature

BlepBlep:
As someone who recently found out about this, I too would really love the ability to adjust (and keep static) the columns in table view. For now I've simply zoomed out enough to see all the columns but this is obviously not ideal lol

kay-zhu:
"torrent title" takes up too much space. I hope it can be dragged to make it bigger or smaller. Thank you.

PR #2042, opened 2024-11-12
fix(TableView): Set max width of torrent name to 40%

Larsluph:
#1401

Tried to have a width editable by the user, I haven't found a way yet.

Quick fix to restrict max width to 40% of the user screen.

### DIFF
--- a/src/components/Dashboard/Views/Table/TableTorrent.vue
+++ b/src/components/Dashboard/Views/Table/TableTorrent.vue
@@ -46,7 +46,7 @@ const getComponent = (type: DashboardPropertyType) => {
 
 <template>
   <template v-for="ppt in torrentProperties">
-    <component v-if="ppt.props" :is="getComponent(ppt.type)" :torrent="torrent" v-bind="ppt.props" />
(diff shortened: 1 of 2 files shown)


WDaan (approved):

puddingsoda:
@Larsluph Would you be able to point me towards making it 0%? Is there any file that I can change? I like my titles hidden most of the time

Larsluph:
@puddingsoda It's a CSS rule, you can override it with an extension.
The tricky part is to apply the same thing on table headers, they don't have targetted classes so you'll have to take advantage of it always being the first one.
Note that this is only applicable for table view.

https://github.com/VueTorrent/VueTorrent/blob/730c930afa9ee2f7b4d274fddfbdfd6af403db90/src/components/Dashboard/Views/Table/TableView.vue#L140-L144

Something like this should make it the smallest it can:

```css
thead > tr > th:nth-child(1) {
  max-width: 0;
  overflow: hidden;
}

#torrentList .torrent-name {
  max-width: 0vw;
}
```

PR #2044, opened 2024-11-13
feat(Logs): Add sort support

### LINKED ISSUE #2035: [Feature Request]: Reverse Log order
### Description

Ideally a toggle, or just a UX thing where the logs are sorted from newest to oldest. (?)

This would permit simpler log viewing as they would now be in a more logical layout than the current `oldest -> newest`. 

Simple `Array.toReversed()`?

_No response_

jagadam97:
# sorting functionality in logs view [feat]

Closes #2035

As requested, I have implemented a sort button in logs view

Screenshot 2024-11-14 at 4 27 40 AM
Screenshot 2024-11-14 at 4 29 30 AM
Screenshot 2024-11-14 at 4 27 40 AM

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/stores/logs.ts
+++ b/src/stores/logs.ts
@@ -26,7 +26,7 @@ export const useLogStore = defineStore(
     const filteredLogsByType = computed(() => logs.value.filter(log => logTypeFilter.value.includes(log.type)))
     const { results: filteredLogs } = useSearchQuery(filteredLogsByType, logMessageFilter, log => log.message)
     const { paginatedResults, currentPage, pageCount } = useArrayPagination(
-      () => filteredLogs.value.toSorted((a, b) => comparators.numeric.compare(a.id, b.id, reverseSort.value)),
+      () => filteredLogs.value.toSorted((a, b) => comparators.numeric.compare(a.id, b.id, !reverseSort.value)),
       30
     )
     const logTask = useTask(function* (_: AbortSignal, lastId?: number) {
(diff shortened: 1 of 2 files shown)


src/stores/logs.ts
```diff
@@ -18,14 +18,21 @@ export const useLogStore = defineStore(
...
     const logTypeFilter = ref<LogType[]>([LogType.NORMAL, LogType.INFO, LogType.WARNING, LogType.CRITICAL])
     const logMessageFilter = ref('')
 
     const filteredLogsByType = computed(() => logs.value.filter(log => logTypeFilter.value.includes(log.type)))
     const { results: filteredLogs } = useSearchQuery(filteredLogsByType, logMessageFilter, log => log.message)
-    const { paginatedResults, currentPage, pageCount } = useArrayPagination(filteredLogs, 30)
+    
+    const finalLogs = computed(() => {
+      return reverseSort.value ? [...filteredLogs.value].reverse() : [...filteredLogs.value];
+    });
```
Larsluph, line 30: Prevent duplicating array if not needed

```suggestion
    const finalLogs = computed(() => reverseSort.value ? filteredLogs.value.toReversed() : filteredLogs.value)
```

Larsluph (changes requested):
Thanks for the contribution!

src/stores/logs.ts
jagadam97: removed duplicate array.

as of using useSorted i am getting into type  mis-match issue  with simple implementation. (dont want to complicate code).

so went with sorting in place. (Hope this will be file )

Larsluph: VueUse composables returns new refs so you have to create a new constant.
You're good with this, I'll push what's needed.

Larsluph (approved):

PR #2076, opened 2024-12-07
Fixed disabling of i2p settings when i2p is enabled

Kladki:
# Fixed disabling of i2p settings when i2p is enabled

Previously, the I2P settings were disabled when I2P itself was disabled and vice versa, leading to the following:
image
image

This PR fixes this, so that you can edit the settings when I2P is enabled, and cannot when it is disabled.

## PR Checklist

- [x] All tests pass


Larsluph (approved):
Awesome, thanks for the fix!

PR #2096, opened 2024-12-18
fix(RSS): Display all feeds even if articles empty

vancez:
Display all feeds even if articles empty for editing or deletion


Larsluph (approved):
Great! Thanks for the contribution!

PR #2100, opened 2024-12-23
fix(Logs): Prevent duplication on page refresh

vancez:
https://github.com/VueTorrent/VueTorrent/blob/098ee25593ff1a483999e3b67bab68dc5a97a278/src/pages/Logs.vue#L75
https://github.com/VueTorrent/VueTorrent/blob/098ee25593ff1a483999e3b67bab68dc5a97a278/src/App.vue#L83


Larsluph (approved):
Thanks for the contribution!

PR #2108, opened 2024-12-29
feat(TorrentDetail): Add torrent navigation buttons

### LINKED ISSUE #2106: [Feature Request]: Arrow buttons to switch between torrents in detail view
### Description

When I'm renaming files I have to get in and out between the torrent list and the detail view.

If I filter and sort i can get a sorted list of the torrents I have to rename, then I can enter into the first torrent detail view, rename the files and then, I should be able to navigate to the next torrent clicking the right arrow from the detail view without exiting.
image

_No response_

Larsluph:
Fixes #2106

This PR adds 4 buttons to the Torrent Detail view. These buttons' actions are as follows from left to right:
- go to first filtered torrent
- go to previous filtered torrent
- go to next filtered torrent
- go to last filtered torrent

image

Also a new store has been introduced to allow for remounting the router component and ensure that lifecycle event are properly executed.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -6,7 +6,7 @@ import Navbar from '@/components/Navbar/Navbar.vue'
 import { TitleOptions } from '@/constants/vuetorrent'
 import { formatPercent, formatSpeed } from '@/helpers'
 import { backend } from '@/services/backend'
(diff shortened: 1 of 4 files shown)


WDaan (approved):

Phil57:
Hello, this works fine for for me it breaks the content tab. No matter the torrent I select, the content tab will always display content of the first torrent of the list.
Other tabs are fine.

Larsluph:
I was just looking into that issue too. It looks likes dependency changes broke some stuff in the reactivity of the app, will run rollback and release patch.

PR #2109, opened 2024-12-30
perf(Dashboard): Make scroll bars always visible

### LINKED ISSUE #2079: [Feature Request]: Make horizontal scroll bar always visible
### Description

Right now the horizontal scroll  bar is only at bottom.

If we can get an option like the pagination to put additional scroll bar at top it would be great. 

_No response_

_No response_

Larsluph:
Fixes #2079

### DIFF
--- a/src/components/Dashboard/Views/Table/TableView.vue
+++ b/src/components/Dashboard/Views/Table/TableView.vue
@@ -77,7 +77,7 @@ const getTorrentRowColorClass = (torrent: TorrentType) => [isTorrentSelected(tor
 
     <template #item="{ item: torrent }">
       <tr
-        :class="['cursor-pointer', 'selected', getTorrentRowColorClass(torrent)]"
+        :class="['cursor-pointer', 'selected', 'ripple-fix', getTorrentRowColorClass(torrent)]"
         v-ripple
         @contextmenu="$emit('onTorrentRightClick', $event, torrent)"
         @touchcancel="$emit('endPress')"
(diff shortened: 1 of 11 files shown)


WDaan (approved):

src/components/Dashboard/Views/Table/TableView.vue
```diff
@@ -46,19 +54,30 @@ const getTorrentRowColorClass = (torrent: TorrentType) => [isTorrentSelected(tor
...
       </tr>
-    </thead>
-    <tbody>
+    </template>
+
+    <template #item="{ item: torrent }">
       <tr
-        v-for="torrent in paginatedTorrents"
-        :class="['cursor-pointer', 'selected', 'ripple-fix', getTorrentRowColorClass(torrent)]"
+        :class="['cursor-pointer', 'selected', getTorrentRowColorClass(torrent)]"
```
asrevni, line 80: This reverts to ripple effect issue found with Safari browser. Any reason for ommiting `ripple-fix`?

Larsluph: I checked the code and bo class with that name exists in our CSS. I don't remember having fixed that issue...

asrevni: #1864. It's still a valid workaround.

Larsluph: Well I may have messed something up while looking for it, thanks for the review!

asrevni: Thank you too!

Issue #2092, opened 2024-12-15
[Feature Request]: (Accessibility) Darker dark theme (i.e. lower white point)


tavern2782:
### Description

I would like to see a more muted / less bright dark theme. I have an OLED monitor and find the colors in the dark theme to be far too bright for my eyes.

### Describe the solution you'd like

One of the two below would be nice to see:

- Settings -> VueTorrent -> General -> Dark Theme Variant -> Redesigned (Low white point)
- Settings -> VueTorrent -> General -> Accessibility -> Lower white point (or maybe "Lower brightness") 

### Describe alternatives you've considered

Up until Dark Reader updated a few days ago on Firefox, I was using this extension to make the VueTorrent dark theme perfect. Something changed in how the extension processes the CSS elements in VueTorrent which now makes all colors completely disappear, making contrast almost non-existent.

### Complementary informations

- [ ] Is this feature already implemented in the default WebUI?

Larsluph:
A new OLED theme seems to be the best way to handle this case, feel free to create a new dark theme that suits your expectations and submit a PR with your changes, some native vuetify colors may not be present in other themes so feel free to ask for guidance in the discord server.

tavern2782:
> A new OLED theme seems to be the best way to handle this case, feel free to create a new dark theme that suits your expectations and submit a PR with your changes, some native vuetify colors may not be present in other themes so feel free to ask for guidance in the discord server.

Hi @Larsluph , thank you for the kind reply. I like your idea.

I've never coded in JS before, but am willing to give it a shot. When I have some time this week from work, I will join the Discord as you suggest so that I can ask for guidance.

tavern2782:
I did some digging into the code, and it seems like the theme isn't the issue. The specific colors that hurt my eyes are defaults and generated colors.

https://github.com/VueTorrent/VueTorrent/blob/dcaa55ebbb553eff0aede10a6a5fdee5366d7875/src/themes/global.ts#L52
The "blue-lighten-1" here for the dark theme really hurts.

The others are generated chip colors on category and tracker, which is much harder to fix. I appreciate the generated colors to provide visual differentiation, whereas the defaults colors do not provide such nice UI. However, there's no easy way to make the generated colors not resolve to a retina burning color on OLED.

Larsluph:
> I've never coded in JS before

Don't worry about that, you'll just have to change some JSON values. Everything else is reloaded automatically!

[...]

PR #2125, opened 2025-01-02
Feat: Add OLED theme

tavern2782:
# Add OLED theme [feat]

https://github.com/VueTorrent/VueTorrent/issues/2092

I've imported functionality from `tinycolor` that darkens colors for OLED monitors. I've found that many of the dynamically generated colors for chips and other items are a bit too bright on OLED.

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Core/ColoredChip.vue
+++ b/src/components/Core/ColoredChip.vue
@@ -21,18 +21,20 @@ const props = withDefaults(
 const { t } = useI18nUtils()
 const { current } = useTheme()
 const { enableHashColors, hideColoredChip } = storeToRefs(useVueTorrentStore())
+const vueTorrentStore = useVueTorrentStore()
 
 function getThemeColor(color: string) {
   return current.value.colors[color] ?? color
 }
 
-// Darken colors for OLED theme
-const vueTorrentStore = useVueTorrentStore()
-let darkenColors = false
-if (vueTorrentStore.theme.mode === 'dark' && vueTorrentStore.theme.dark === 'dark-oled') darkenColors = true
+function transformColor(color: TinyColor) {
+  if (isDarkOledTheme.value) return color.darken(30)
+  else return color
+}
 
-const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value, darkenColors)))
-const rawChipColor = computed(() => (props.disabled || !enableHashColors.value ? getThemeColor(props.defaultColor) : getColorFromName(props.value, darkenColors)))
+const isDarkOledTheme = computed(() => vueTorrentStore.theme.mode === 'dark' && vueTorrentStore.theme.dark === 'dark-oled')
+const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value, transformColor)))
+const rawChipColor = computed(() => (props.disabled || !enableHashColors.value ? getThemeColor(props.defaultColor) : getColorFromName(props.value, transformColor)))
 const chipValue = computed(() => (props.disabled ? props.disabledValue || props.value || t('common.none') : props.value))
 const shouldShowColoredChips = computed(() => !hideColoredChip.value)
 </script>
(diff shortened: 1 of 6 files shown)


src/components/Core/ColoredChip.vue
```diff
@@ -26,8 +26,13 @@ function getThemeColor(color: string) {
...
 
-const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value)))
-const rawChipColor = computed(() => (props.disabled || !enableHashColors.value ? getThemeColor(props.defaultColor) : getColorFromName(props.value)))
+// Darken colors for OLED theme
+const vueTorrentStore = useVueTorrentStore()
+let darkenColors = false
+if (vueTorrentStore.theme.mode === 'dark' && vueTorrentStore.theme.dark === 'dark-oled') darkenColors = true
+
+const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value, darkenColors)))
+const rawChipColor = computed(() => (props.disabled || !enableHashColors.value ? getThemeColor(props.defaultColor) : getColorFromName(props.value, darkenColors)))
```
Larsluph, line 35: Convert this to a computed property to conserve reactivity (when changing themes)

```suggestion
const isDarkOledTheme = computed(() => vueTorrentStore.theme.mode === 'dark' && vueTorrentStore.theme.dark === 'dark-oled')
const chipColor = computed(() => (props.disabled || !enableHashColors.value ? props.defaultColor : getColorFromName(props.value, isDarkOledTheme.value)))
const rawChipColor = computed(() => (props.disabled || !enableHashColors.value ? getThemeColor(props.defaultColor) : getColorFromName(props.value, isDarkOledTheme.value)))
```

src/helpers/colors.ts
```diff
@@ -9,12 +9,13 @@ function djb2Hash(str: string): number {
   return hash >>> 0 // ensure non-negative integer
 }
 
-export function getColorFromName(name: string) {
+export function getColorFromName(name: string, darkenColors?: boolean) {
```
Larsluph, line 12: Now that I see it, it might be better to use a transformer instead of hardcoding this in the helpers. Using this instead will keep everything in their place, and allow for better customization.

```suggestion
export function getColorFromName(name: string, transform?: (color: TinyColor) => TinyColor) {
```

src/themes/dark/oled.ts
```diff
@@ -0,0 +1,19 @@
...
+    colors: {
+      primary: '#35495E',
+      secondary: '#415c75',
+      navbar: '#273845',
+      download: '#5BB974',
+      background: '#121212',
+      selected: colors.grey.darken1,
+      red: colors.red.accent3,
+      ...getVariables(true)
+    }
```
Larsluph, line 17: This object can be imported from the other theme to prevent duplication. You can then override the `torrent-ul_stalled` color with a more color suitable for OLED

tavern2782:
All tests pass and all commented code has been removed. Ready for review :)

tavern2782:
I just noticed something that you may would like to consider for the future.

It appears that the dynamic colors also apply to the text variant. My work here has been targeting just the chips. When testing these changes with the text variant, the colors are too dark and hard to read in the dark theme.

If there is a way to target text variant separate from chips, we may want to consider that for a future PR.

tavern2782:
Deleted my last message because I can live with this. One of the dynamic colors really is a bit too dark in text variant, but it's going to be a game of diminishing returns to keep chasing edge cases.

image

src/components/Core/ColoredChip.vue
```diff
@@ -21,18 +21,20 @@ const props = withDefaults(
 const { t } = useI18nUtils()
 const { current } = useTheme()
 const { enableHashColors, hideColoredChip } = storeToRefs(useVueTorrentStore())
+const vueTorrentStore = useVueTorrentStore()
```
Larsluph, line 24: Prevent duplicate calls
```suggestion
const vueTorrentStore = useVueTorrentStore()
const { enableHashColors, hideColoredChip } = storeToRefs(vueTorrentStore)
```

Larsluph (approved):

PR #2164, opened 2025-02-12
chore(deps): Rollback @vue/eslint-config-typescript to 14.2.0

Larsluph:
Bump caused lint command to fail


Larsluph:
@dependabot ignore @vue/eslint-config-typescript

PR #2253, opened 2025-04-23
feat: Add confirmation dialogs

### LINKED ISSUE #2247: [Feature Request]: Add a confirmation window to ban a peer.
### Description

Sometimes on mobile I like checking to see what peers I'm connected too, but I seem to accidentally ban peers without knowing; so I feel like maybe could add a confirmation popup window to confirm that you really want to ban that peer.

Popup window to confirm that you really want to ban that peer.

_No response_

Larsluph:
Fixes #2247

- RSS Feed deletion
- RSS Rule deletion
- Peer ban
- Tag / Category deletion
- Right Click > remove unused tags / categories


WDaan (approved):
🚀

--- Except for the type checks failing 😓

PR #2369, opened 2025-07-04
fix: PWA app respects Device Rotation lock

jagadam97:
# PWA app respects Device Rotation lock [fix]

PWA app ignores the devices app orientation and rotates when ever i tilt my phone even though the device is locked.
this PR fixes that issue by removing orientation from webmanifest file and let the Underlying browser handle the Rotation which respects the Device settings. will be attaching POT as comment

## PR Checklist

- [x] All tests pass

### DIFF
--- a/public/manifest.webmanifest
+++ b/public/manifest.webmanifest
@@ -35,7 +35,6 @@
     "client_mode": "focus-existing"
   },
   "name": "VueTorrent",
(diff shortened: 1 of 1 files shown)


jagadam97:
https://youtube.com/shorts/ROFuJWZ2eN0?feature=share

Issue #2353, opened 2025-06-23
[Feature Request]: Cache Stats


overlike:
### Description

![Image](https://github.com/user-attachments/assets/22e7848b-f761-4c01-a8ad-1da319b92025)

And as an option, there should be an on/off button.

### Complementary informations

- [ ] Is this feature already implemented in the default WebUI?

overlike:
> Thanks for the request. Data is already fetched and stored [here](https://github.com/VueTorrent/VueTorrent/blob/0c042a3ba70efa4848722de07abb2b58fc4a5964/src/stores/maindata.ts#L19). I guess we can also include all other stats as well.
> 
> I'm not sure from a UX standpoint how we should display it, I feel like the sidebar might get overloaded with data, should we reuse the connection dialog for this? Create another dialog next to it?

should we reuse the connection dialog for this?
No.

Create another dialog next to it?
Yes.

overlike:
> I'm just wondering what's your use case for this kind of data?
> 
> Is it only one-time checks? => Dialog is fine to check values once in a while Is it for monitoring? => Data should be visible right away in the sidebar other cases that might help in the design?

for monitor => Data should be visible right away in the sidebar
And make it optional toggled on or off, in case someone doesn’t want it
Sorry, I'm not good at English. I use translate to answer your question.

overlike:
> In that case, I guess it's time to rework design of those cards, we can probably stack cards side-by-side, and/or with icons?
> 
> Existing layout could look like that:
> 
> ![Image](https://github.com/user-attachments/assets/cb52c740-5fd7-4a20-99a8-8b385787a008)

I think it looks good, it gives more space.

overlike:
I think this is okay.

https://github.com/VueTorrent/VueTorrent/pull/2365

jagadam97:
> I think this is okay. So you don't have to redesign the entire sidebar.
> 
> [#2365](https://github.com/VueTorrent/VueTorrent/pull/2365)

anyway its better to redesign as it takes up so much space in nav bar 
and while at it add the cache status to the side bar which will be option toggle from settings.

PR #2371, opened 2025-07-06
feat(Navbar): More compact layout

Larsluph:
#2353

First part to compact existing layout is done, second part is to inject cache stats from #2365. Still need to figure out which icons to use for all those new stats, I haven't found proper icons yet.

Note: While updating layout, SpeedGraph did some strange things, not sure what broke in the process...

<table>
<tr>
<td>Old</td>
<td>New</td>
</tr>
<tr>
<td>
New
</td>
<td>
New
</td>
</tr>
</table>


WDaan (approved):
Nice! 😎

Issue #2097, opened 2024-12-21
[Bug]: "Checking current auth status..." now showing repeatedly


moracha:
### Bug description

After upgrading from 2.18.0 to 2.19.0, on mobile only (desktop unaffected) I'm now getting a pop-up "Checking current auth status..." at the bottom of Firefox. I've never seen this before and haven't changed any settings, so I'm assuming something changed with this release. I have local subnet whitelist enabled in QB client so I typically never see anything auth related.

Last time I got this pop-up I refreshed the page but it didn't go away. Switching to "desktop site" doesn't help either. Only way to get rid of the pop-up was opening a new browser window.

New window is typically fine until the following day, when the pop-up returns and I have to close and open a new window again. So maybe something on the backend is timing out? Also could be a change in QB 5.0.3 which I recently upgraded to.

![Screenshot_20241220_193857_Firefox](https://github.com/user-attachments/assets/50e1b6fd-aa5c-48fd-a337-9e96e58a8998)

### Expectations

No auth pop-up

### Relevant debug informations

Not sure how to pull debug info on mobile

### Steps to reproduce

1. Access UI on Android Firefox 
2. Wait 12-24h or so
3. "Checking current auth status..." pop-up becomes persistent

### How did you install VueTorrent?

Downloaded from GitHub releases

### Environment description

```plain text
Device type: Phone
OS: Android
Browser (name and version): Firefox 133.0.3

PWA: ?
Running inside a container: No
```

### qBittorrent version

5.0.3

### VueTorrent version

2.19.0

### Agreements

- [X] The bug still occurs on the nightly build
- [X] The bug still occurs on the latest version of qBittorrent
- [X] A similar issue doesn't already exist
- [X] I checked the DevTools and included any errors that may show up

Larsluph:
I also have this bug in dev mode but haven't found any trace in a release.

It usually happens when you reuse an existing session, which has cached information. Opening a new browser window resets that cache so the popup doesn't show up.

moracha:
K, might try clearing browser cache first and then reopening a new window to see if it persists. Will report back if it keeps happening.

Larsluph:
Tried to reproduce and even in dev mode I didn't succeed. It must be some kind of weird edge case that only happens in unexpected scenario.

dillbyrne:
[...]

PR #2509, opened 2025-09-21
fix: "Checking auth status" toast

Larsluph:
#2097

I hope it's enough now.

localhost qbittorrent responded in 46ms, tested with 1ms with mock and it still works.


WDaan (approved):

ebkalderon:
Thanks so much for addressing this so quickly, @Larsluph! You're awesome. I'm hopeful that this truly fixes #2097. :crossed_fingers:

PR #2610, opened 2026-01-04
fix(TorrentCreatorForm): set default private field to true

y3nd:
Setting private flag to true at torrent creation is safer and avoids mis-announcing on DHT/PEX.

## PR Checklist

- [x] All tests pass


Larsluph:
If the default doesn't fit everyone, I guess we may add some kind of memory, synced in the local storage and backend.

local storage example:
https://github.com/VueTorrent/VueTorrent/blob/bbbe846295f4bb23f734fc3595891acac873374a/src/stores/vuetorrent.ts#L318-L323
Backend sync example:
https://github.com/VueTorrent/VueTorrent/blob/bbbe846295f4bb23f734fc3595891acac873374a/src/App.vue#L44-L55

Larsluph:
Closed for inactivity

PR #2721, opened 2026-04-04
Table column resize

### LINKED ISSUE #1401: [Feature Request] Ability to resize columns in table view.

chesterjazzman:
# Table column resize [feat]

Implements #1401 

- Adds a new directive for making a table's column widths sizable.    
- Adds a `<colgroup> `for the table, each of the `<col>` tags, and an invisible resize handle on each `<th>` except for the first little tool tip column.
- Adds mouse listeners for the resize handles to set the col width.
- Adds double-click handler to auto resize width.
- Persists column widths across sessions in the vue torrent pinia store.  
- Adds Unit tests for the new directive.

## PR Checklist

- [x] All tests pass


src/directives/resizableColumns.ts
```diff
@@ -0,0 +1,280 @@
...
+        col.dataset.resizableKey = columnKeys[i]
+      } else {
+        delete col.dataset.resizableKey
+      }
+    })
+
+    resizeObserver?.observe(el)
+
+    // Start at index 1 — the first column (tooltip) is intentionally not resizable.
+    for (let index = 1; index < headerCells.length; index++) {
```
Larsluph: Don't hardcode rules on directives, we might use it on another table which doesn't have the status tooltip column.

Larsluph (commented):
Thanks for the contribution, this is a useful feature and definitely something the community asked!

What led you to implement this as a directive rather than introducing a dedicated table component?

From what I see, the directive currently includes some logic that is tightly coupled to a specific table (hardcoded rule). That makes me wonder if we’re starting to mix generic behavior (column resizing) with `TableView`-specific concerns.

In the long run, this might make it harder to:
- reuse the feature across different tables
- evolve it with additional capabilities (e.g. column reordering)
- keep the logic maintainable as more features get added

Given that we already anticipate more table-related features, would you be open to exploring a component-based approach instead (e.g. a dedicated resizable/dynamic table)? That would give us a clearer place to centralize state and extend functionality over time.

Larsluph:
> I have had bad experiences on other projects when I suggest altering dependencies.

We've seen a surge of issues and hacks about NPM dependencies, especially lately. We must be very cautious about which dependency we allow/use.

> Are you suggesting a general table component that builds on top of v-data-table, or replacing it entirely?

I would say it really depends on how you will implement it. I would prefereably opt for a layer on top of the existing one. But if that makes it unstable or "hacky", we are not closed to exploring other suggestions.

> If the latter, it might be simpler to use one of these?

Only if it can be adapted on the existing Vuetify layout (i.e. only importing the component code, replace styling in accordance with Vuetify). It's not a good idea to use several component library on the same project.

> I might also suggest that with additional changes to this directive, it could play nicely with another draggableColumn directive. Not sure what other plans you have.

There's a label for all `TableView` requests: [is:issue state:open label:"Table View"](https://github.com/VueTorrent/VueTorrent/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22Table%20View%22)

But basically:
- column freeze
- column resize
- column reordering

chesterjazzman:
I started all this before reading your latest comment.  But it looks like this meets what you prefer.

src/components/Dashboard/Views/Table/TableView.vue
```diff
@@ -148,5 +149,11 @@ function getTorrentRowColorClass(torrent: TorrentType) {
     overflow: hidden;
     text-overflow: ellipsis;
   }
+  th,
+  td {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
```
Larsluph: I guess these can be included in `TableComponent`?

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,252 @@
...
+}
+
+function clearResizeHandles(root: HTMLElement) {
+  while (resizeHandleCleanups.length) {
+    resizeHandleCleanups.pop()?.()
+  }
+  root.querySelectorAll(`.${RESIZE_HANDLE_CLASS}`).forEach(node => node.remove())
+}
+
+function drawResizeHandles() {
```
Larsluph: This function is quite huge, please simplify into smaller functions.

Don't hesitate to create a helper file with pure function logic.

WDaan:
@CodeRabbit review

src/components/Dashboard/Views/Table/TableView.vue
chesterjazzman: fixed

src/components/Core/TableComponent.vue
chesterjazzman: split it out within the component, and moved pure functions to helpers/tableResize.ts

src/components/Core/TableComponent.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+  return {
+    useVueTorrentStore: () => ({
+      tableColumnWidths,
+      setTableColumnWidth: (tableKey: string, columnKey: string, width: number) => {
+        if (!tableKey || !columnKey || !Number.isFinite(width) || width <= 0) return
+        if (!tableColumnWidths[tableKey]) {
+          tableColumnWidths[tableKey] = {}
+        }
+        tableColumnWidths[tableKey][columnKey] = Math.ceil(width)
+      },
```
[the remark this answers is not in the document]
chesterjazzman: fixed

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,263 @@
...
+        if (resizeTableKey && columnKey) {
+          const width = getResizeColumnWidth(th, currentCol)
+          vuetorrentStore.setTableColumnWidth(resizeTableKey, columnKey, width)
+        }
+        queueResizeRedraw()
+      }
+
+      document.addEventListener('mousemove', onResizeMouseMove)
+      document.addEventListener('mouseup', onResizeMouseUp)
+    }
```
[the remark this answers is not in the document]
chesterjazzman: this seems like a sufficiently rare edge case to go unfixed.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,263 @@
...
+
+<style lang="scss">
+@use 'vuetify/settings';
+
+th,
+td {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
```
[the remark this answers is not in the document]
chesterjazzman: fixed.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,181 @@
...
+        if (resizeTableKey && columnKey) {
+          vuetorrentStore.clearTableColumnWidth(resizeTableKey, columnKey)
+        }
+        requestAnimationFrame(() => {
+          const naturalWidth = Math.max(th.getBoundingClientRect().width, MIN_RESIZE_COLUMN_WIDTH)
+          table.style.tableLayout = 'fixed'
+          setResizeColumnWidth(currentCol, th, naturalWidth)
+          resizeObserver?.observe(resizeRoot)
+        })
+        return
```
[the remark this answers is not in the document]
chesterjazzman: I tried this early on in development.  using a standard double-click listener alongside a mouse-down listener doesn't work.  mouse down triggers twice and double-click never fires.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,181 @@
...
+
+<style lang="scss" scoped>
+@use 'vuetify/settings';
+
+th,
+td {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
```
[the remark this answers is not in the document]
chesterjazzman: fixed

src/helpers/tableResize.ts
```diff
@@ -0,0 +1,101 @@
...
+  const widths = headerCells.map((th, i) => {
+    const key = columnKeys[i]
+    const col = resizeColumns[i]
+    const isSameKey = key ? col.dataset.resizableKey === key : col.dataset.resizableKey === undefined
+    const existing = Number.parseFloat(col.style.width)
+    if (isSameKey && Number.isFinite(existing) && existing > 0) return existing
+    const persistedWidth = key ? persistedWidths[key] : undefined
+    if (Number.isFinite(persistedWidth) && (persistedWidth as number) > 0) return persistedWidth as number
+    return th.getBoundingClientRect().width
+  })
```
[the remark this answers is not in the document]
chesterjazzman: I can't imagine how this situation would occur.  the minimum was in place before it was stored in the first place.  the stored value should never be below the minimum.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,181 @@
...
+
+    const resizeHandle = createResizeHandle()
+
+    function onResizeMouseDown(event: MouseEvent) {
+      const now = Date.now()
+      const isDoubleClick = now - (resizeLastMouseDownTimes.get(index) ?? 0) < 300
+      resizeLastMouseDownTimes.set(index, now)
+
+      event.preventDefault()
+      event.stopPropagation()
```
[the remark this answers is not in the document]
chesterjazzman: this is false.  click() does not fire on the containing th when clicking the resize handler.  verified with testing.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,181 @@
...
+        th.style.width = ''
+        if (resizeTableKey && columnKey) {
+          vuetorrentStore.clearTableColumnWidth(resizeTableKey, columnKey)
+        }
+        requestAnimationFrame(() => {
+          const naturalWidth = Math.max(th.getBoundingClientRect().width, MIN_RESIZE_COLUMN_WIDTH)
+          table.style.tableLayout = 'fixed'
+          setResizeColumnWidth(currentCol, th, naturalWidth)
+          resizeObserver?.observe(resizeRoot)
+        })
```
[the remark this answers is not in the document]
chesterjazzman: fixed

src/helpers/tableResize.ts
Larsluph: Duplicated logic here, we can extract into clearer functions.

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,183 @@
...
+
+      document.addEventListener('mousemove', onResizeMouseMove)
+      document.addEventListener('mouseup', onResizeMouseUp)
+    }
+
+    resizeHandle.addEventListener('mousedown', onResizeMouseDown)
+    resizeHandleCleanups.push(() => resizeHandle.removeEventListener('mousedown', onResizeMouseDown))
+    th.appendChild(resizeHandle)
+  }
+}
```
Larsluph: I feel like this part (at least event management) can be extracted to a composable to simplify readability

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,183 @@
...
+    <v-data-table v-bind="tableAttrs as any">
+      <template v-for="name in slotNames" #[name]="slotProps">
+        <slot :name="name" v-bind="slotProps" />
+      </template>
+    </v-data-table>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+@use 'vuetify/settings';
```
Larsluph: Is it still used?

src/helpers/tableResize.ts
chesterjazzman: broke down into column width helpers.  fixed.

src/components/Core/TableComponent.vue
chesterjazzman: moved to src/composables/TableResize.ts

src/components/Core/TableComponent.vue
chesterjazzman: nope.  good catch.  copy/paste issue.  I removed it.

Larsluph (dismissed):

chesterjazzman:
@Larsluph asked about a visual indicator for the resize handle.  I've started looking into this.  It's going to be another refactor at this point.  I can't get it to look and behave right without changing the DOM structure of the resize handles.  Making it invisible hid slight alignment issues that must be fixed now that the handles are visible.  

My plans are to refactor the code so the handles are children of an overlay div that sits on top of the headers.  That way the `<th>` elements don't own their own handles.  The headers owning the handles caused awkwardness either in cursor pointer targeting, or visual alignment not looking quite right.  When the handle is pushed outside the `<th>` to live right on the seam of the two columns, the cursor targeting stopped working when half of it lived outside the owning `<th>`.  When the whole handle is moved into the `<th>` the alignment looks off, because it is.  The latter is where I've left the branch.  At least this way you can drag slightly to the left and the right of the visual indicator.  

I'll start the refactor to grant the best of both worlds while keeping none of the bad side effects.  Once the `<th>` doesn't own the handle, targeting it when it's directly on the seam becomes a non-issue.  Wiring it to the right column becomes a property issue instead of a parent-child relationship.  And positioning will need to be calculated.

chesterjazzman:
Actually...  no.  after reviewing the default qBit web UI again, I confirmed their resize handles also only extend into the column they are resizing.  You can't grab to the right of the handle to resize the column to the left.  So I'm keeping this to match.  Aligning the seams directly on the `<th>` border, and only allowing extra invisible handle width on the left side.  Ready for review.  @Larsluph

src/helpers/tableResize.ts
```diff
@@ -0,0 +1,104 @@
...
+  Object.assign(resizeHandle.style, {
+    position: 'absolute',
+    top: '0',
+    right: '0',
+    width: `${RESIZE_HANDLE_WIDTH}px`,
+    height: '100%',
+    cursor: 'col-resize',
+    userSelect: 'none',
+    touchAction: 'none',
+  })
```
Larsluph: Use a CSS class instead of duplicating inline style for better control (debug and user-style)

src/components/Core/TableComponent.vue
```diff
@@ -0,0 +1,60 @@
...
+
+        &::after {
+          content: '';
+          position: absolute;
+          top: 18%;
+          bottom: 18%;
+          right: 0;
+          width: 1px;
+          background: rgba(var(--v-theme-on-surface), 0.25);
+        }
```
Larsluph, line 66: Using `::after` for styling the handle will display it next to the real drag zone. It felt off during testing.

Larsluph: Just read your comment back and I got what you meant, I feel like we can still improve it by adding margins on left and right of the element?

chesterjazzman: I don't think there's a real difference between fixed width and margin here.  The column `<th>` next to it on the right wins the mouse-over fight every time.  Same "issue/quirk" on qbit default UI.  

I thought about keeping the CSS such that it looks like it's trying to have left and right area to drag, but since that doesn't actually work, I felt being explicit about only making the left side draggable was more clear.  

And the other way, where I put the seam in the center of the handle and keep the entire handle in the `<th>` so the right side handle is targetable just looks weird.  Here's a screenshot:  
image

src/helpers/tableResize.ts
chesterjazzman: fixed.

src/components/Core/TableComponent.vue
chesterjazzman: Here's a screenshot of the way that looks nice, but isn't targetable on the right side of the line: 
image

src/components/Core/TableComponent.vue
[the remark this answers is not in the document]
chesterjazzman: this is actually fixed now.  The original behavior of mouse down preventing double-click was because there was an infinite build/tear down loop in the original design.  Essentially, before the user had a chance to click twice the element was destroyed and recreated.  This has been fixed, and now a real double-click listener is working.  fixed on this commit: https://github.com/VueTorrent/VueTorrent/pull/2721/commits/b2b81752278cb8b9bfa820719d4c468c265974a1 and https://github.com/VueTorrent/VueTorrent/pull/2721/commits/009aed7a29ee83e26222191874f4afa243ea1d99

src/components/Core/TableComponent.vue
Larsluph: > And the other way, where I put the seam in the center of the handle and keep the entire handle in the <th> so the right side handle is targetable just looks weird.

I think it looks weird because the text is not aligned properly, maybe by adjusting left padding of the text cell.

chesterjazzman: well, okay.  the resize handle is 12px.  So I subtracted 6px from the default left padding, and added 6px to the default right padding.  That shifted everything over and the resize handle is targettable from both sides.  
image

Larsluph (approved):

PR #2775, opened 2026-05-11
fix(login): restore vertical centering after Vuetify v4 migration

### LINKED ISSUE #2770: [Bug]: Login Screen layout regression after Vuetify migration
### Bug description

It also looks ugly on mobile too.

On v2.33.0
Image

### Expectations

Look like before v2.33.0

Image

n/a

### Steps to reproduce

n/a

### Environment description

```plain text
Device type: Desktop
OS: Windows 11
Browser: 147.0.3912.98 (Official build) (64-bit)
```

### qBittorrent version

5.2.0

### VueTorrent version

v2.33.0

### Agreements

Akur1a:
## Summary

- Restores vertical centering of the login card after the Vuetify v4 migration
- Vuetify 4 removed the `.v-container.fill-height` CSS rule that previously added `display: flex; align-items: center`
- Added explicit Vuetify utility classes (`d-flex align-center justify-center`) to restore the expected centered layout

## Test plan

- [ ] Open the login page on desktop — card should be vertically and horizontally centered
- [ ] Open the login page on mobile — card should be centered, not stuck at the top
- [ ] Verify login functionality still works (username/password fields, submit button)

Fixes #2770

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/components/Core/AutofillableField.vue
+++ b/src/components/Core/AutofillableField.vue
@@ -96,7 +96,6 @@ input {
   -webkit-appearance: none;
   background: transparent;
   border: 0;
-  border-radius: 0;
   box-shadow: none;
   color: inherit;
   min-width: 0;
@@ -108,15 +107,5 @@ input:focus-visible,
 input:active {
   box-shadow: none;
   outline: none;
-}
-
-input:-webkit-autofill,
-input:-webkit-autofill:hover,
-input:-webkit-autofill:focus,
-input:-webkit-autofill:active {
-  -webkit-text-fill-color: currentColor;
-  caret-color: currentColor;
-  box-shadow: 0 0 0 1000px transparent inset;
-  transition: background-color 9999s ease-in-out 0s;
 }
 </style>
--- a/src/components/Core/AutofillableField.vue
+++ b/src/components/Core/AutofillableField.vue
@@ -94,4 +94,10 @@ div.autofill-container--disabled {
 input:focus {
   outline: none;
 }
+
+input {
+  background: transparent;
+  border: 0;
+  color: inherit;
+}
 </style>
--- a/src/components/Core/AutofillableField.vue
+++ b/src/components/Core/AutofillableField.vue
@@ -91,13 +91,32 @@ div.autofill-container--disabled {
   pointer-events: none;
 }
 
-input:focus {
+input {
+  appearance: none;
+  -webkit-appearance: none;
+  background: transparent;
+  border: 0;
+  border-radius: 0;
+  box-shadow: none;
+  color: inherit;
+  min-width: 0;
   outline: none;
 }
 
-input {
-  background: transparent;
-  border: 0;
-  color: inherit;
+input:focus,
+input:focus-visible,
+input:active {
+  box-shadow: none;
+  outline: none;
+}
+
+input:-webkit-autofill,
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus,
+input:-webkit-autofill:active {
+  -webkit-text-fill-color: currentColor;
+  caret-color: currentColor;
+  box-shadow: 0 0 0 1000px transparent inset;
+  transition: background-color 9999s ease-in-out 0s;
 }
 </style>
--- a/src/pages/Login.vue
+++ b/src/pages/Login.vue
@@ -54,7 +54,7 @@ watchEffect(() => {
 </script>
 
 <template>
-  <v-container class="fill-height">
+  <v-container class="fill-height d-flex align-center justify-center">
     <v-card class="mx-auto" rounded="lg" min-width="250">
       <v-card-title>{{ t('login.title') }}</v-card-title>
       <v-card-subtitle>{{ t('login.subtitle') }}</v-card-subtitle>


Larsluph:
Please provide screenshots of changes

Akur1a:
Added screenshots of the login page after the layout fix.

Desktop (1365x768):

Desktop login screenshot

Mobile (390x844):

Mobile login screenshot

Also verified locally that the login card is centered in both viewports, and `npm run lint -- src/pages/Login.vue` passes.

Larsluph:
Can you also investigate why the text field has the built-in border? It shouldn't be there from the previous version screenshot.

Akur1a:
Thanks, I investigated this as well.

The visible black border was not introduced by the centering change in this PR. That change only adds flex alignment classes to the login page container. The border came from the native `<input>` inside the custom `AutofillableField` component: after the Vuetify 4 upgrade, the browser default input border was no longer reset there, so it became visible on the login fields.

I updated `AutofillableField` to reset the native input border/background while keeping the existing custom field wrapper styling intact.

Verified with:
- `npm run lint -- src/pages/Login.vue src/components/Core/AutofillableField.vue src/components/Core/AutofillableField.spec.ts`
- `npm run test -- src/components/Core/AutofillableField.spec.ts`

Updated screenshots:

Desktop (1365x768):

Desktop login screenshot

Mobile (390x844):

Mobile login screenshot

src/components/Core/AutofillableField.vue
```diff
@@ -94,4 +94,10 @@ div.autofill-container--disabled {
 input:focus {
   outline: none;
 }
+
+input {
+  background: transparent;
+  border: 0;
+  color: inherit;
```
jagadam97, line 100: you dont need color right?? can you check this.

jagadam97:
also can yoyu look into these 

image

image

src/components/Core/AutofillableField.vue
Akur1a: I kept `color: inherit` intentionally so the native input text follows the surrounding Vuetify theme color in both light and dark themes. The visual issue here comes from the browser's native input chrome, so the new reset targets appearance, box-shadow, outline, and border radius while preserving the inherited text color.

Akur1a:
﻿Thanks, I updated the shared `AutofillableField` reset for this.

I also checked the other usages of `AutofillableField` / `PasswordField`. The same native input chrome can show up in the login fields and in settings fields such as Connection proxy auth and Web UI authentication, because they all use the same shared component.

This is technically beyond the original "login page centering" scope of this PR, but it addresses the input border issue raised in the review. The change is scoped to `AutofillableField` so the existing outer rounded field border/focus styling stays intact while the inner native input border/shadow/appearance is removed consistently.

jagadam97:
> ﻿Thanks, I updated the shared `AutofillableField` reset for this.
> 
> I also checked the other usages of `AutofillableField` / `PasswordField`. The same native input chrome can show up in the login fields and in settings fields such as Connection proxy auth and Web UI authentication, because they all use the same shared component.
> 
> This is technically beyond the original "login page centering" scope of this PR, but it addresses the input border issue raised in the review. The change is scoped to `AutofillableField` so the existing outer rounded field border/focus styling stays intact while the inner native input border/shadow/appearance is removed consistently.
> 


Since the fix lives in the shared `AutofillableField` component, it naturally regresses the same bug everywhere it's used — login, proxy auth, and Web UI auth. Rather than opening a follow-up PR for issues this change already addresses, would you be open to expanding the scope here to cover all instances? Feels cleaner to land them together.

Akur1a:
﻿Yes, I checked the other usages as well. The custom native input is centralized in `AutofillableField`, and `PasswordField` wraps the same component, so this reset should cover the affected login/settings fields consistently.

I did not find another separate raw input path with the same issue. I'm fine with keeping this in this PR if the maintainers are okay with the expanded scope.

jagadam97:
@Larsluph any insight on this??

Larsluph:
We can just change this PR to a styling fix PR instead. I really don't care about the scope of any change, as long as it stays about the same general topic, especially here with a major Vuetify migration.

@jagadam97 You opened a PR with only a few style changes. Does these fix the issue in the same way? I feel like adding more styles override will only complexify future maintenance.

Akur1a:
@Larsluph @jagadam97 I took another look at this after comparing it with #2783.

I think the previous reset was a bit broader than it needed to be, especially the WebKit autofill part. I trimmed that back so we are not adding more long-term styling override than necessary.

Compared with #2783, this still keeps a couple of extra native input resets for browser appearance, focus/active states, and box-shadow. So it is a little more defensive than just `background` + `border`, but much smaller than the previous version and should still cover the shared `AutofillableField` / `PasswordField` cases.

jagadam97:
> @jagadam97 You opened a PR with only a few style changes. Does these fix the issue in the same way?

yes

> feel like adding more styles override will only complexify future maintenance.

agreeed but i have not found a better fix.

WDaan (approved):

Larsluph (approved):

PR #2849, opened 2026-07-04
Add useFilenameAsRename option for adding torrents [feat]

chocolate-croissant:
This PR introduces a new toggle in the Add Torrent dialog to automatically use the uploaded torrent's filename as the initial value for the rename field.

When adding a torrent file, checking the "Use Filename" option will strip the `.torrent` extension and populate the "Rename" input, saving the user from manually typing out the name.

**Motivation:**
This is especially useful when handling large batches of torrents (e.g., `1.torrent`, `2.torrent`) that the user has already pre-organized by renaming the `.torrent` files directly prior to uploading them to the UI.

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Dialogs/AddTorrentDialog.vue
+++ b/src/components/Dialogs/AddTorrentDialog.vue
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
 import { storeToRefs } from 'pinia'
-import { computed, onBeforeMount, ref } from 'vue'
+import { computed, onBeforeMount, ref, watch } from 'vue'
 import { toast } from 'vue3-toastify'
 import AddTorrentParamsForm from './AddTorrentParamsForm.vue'
 import HistoryField from '@/components/Core/HistoryField.vue'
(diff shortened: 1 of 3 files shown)


Larsluph (approved):

Larsluph (changes requested):
Lint doesn't pass and nitpick comment from coderabbit is a nice-to-have.

chocolate-croissant:
Thanks for the thorough review!

I've pushed a new commit that addresses the nitpick by extracting the regex replacement into a shared helper function, and completely refactors the submission loop for this feature:
1. Switched to `Promise.allSettled` to ensure we wait for all uploads to complete, regardless of individual success/failure.
2. Individual file promises now explicitly track their respective `file.name` so that if they fail, we can isolate them. If partial (or total) failures occur, we now format a highly detailed error message (e.g., `3/10 torrent(s) failed to add. Failed torrents: 1.torrent, 2.torrent, 3.torrent`) and throw it to be handled by the toast UI.
3. Any files that successfully upload are now removed from the internal file selection state. This ensures that if errors occur and the dialog stays open, it only contains the files that actually failed, preventing users from accidentally duplicating their successful uploads on a retry.

src/components/Dialogs/AddTorrentDialog.vue
```diff
@@ -73,12 +87,68 @@ function submit() {
...
+    for (const file of files.value) {
+      const filePayload = { ...payload, rename: stripTorrentExtension(file.name) }
+      const p = torrentStore
+        .addTorrents([file], '', filePayload)
+        .then(() => {
+          files.value = files.value.filter(f => f.name !== file.name)
+        })
+        .catch(err => {
+          const e = new Error(err.message || String(err))
+          ;(e as any).fileName = file.name
```
Larsluph, line 103: This can be extracted in a new Error subclass for better typing

Larsluph (changes requested):

PR #2860, opened 2026-07-20
feat(search): Configurable default sort order for search results

### LINKED ISSUE #2859: [Feature Request]: Default sort orders in Search Engine
### Description

I always want the search engine results to default to torrents with the most seeders at the top. Other people will want different defaults, great.

A way of default ordering columns in the search engine.

_No response_

samcolson4:
Closes #2859

## Summary
- Adds a **Search Engine** subtab under Settings → VueTorrent to configure a default sort column/order for search-result tables, since qBittorrent's API has no server-side preference for this.
- New search tabs are seeded with the configured default sort; changing the sort on an individual tab (via the table header) only affects that tab, not the saved default.
- Adds a `SearchEngineColumn` enum shared between the settings UI and the search-result table sort state, filtered by `appStore.usesQbit5` to match the column logic already used on the Search Engine page.

## Test plan
- [x] `npm run lint`
- [x] `npm run test` (376 passed)
- [X] Manually verify in the browser: set a default sort in Settings → VueTorrent → Search Engine, open a new search tab, confirm it's pre-sorted; change sort on the tab and confirm the saved default is unaffected.

## Screenshots
Screenshot 2026-07-16 at 09 30 45

Screenshot 2026-07-16 at 09 30 57

### DIFF
--- a/src/components/Settings/VueTorrent/SearchEngine.vue
+++ b/src/components/Settings/VueTorrent/SearchEngine.vue
@@ -1,0 +1,55 @@
+<script setup lang="ts">
+import { computed } from 'vue'
+import { useI18nUtils } from '@/composables'
+import { SearchEngineColumn } from '@/constants/vuetorrent'
+import { useAppStore, useVueTorrentStore } from '@/stores'
+
+const { t } = useI18nUtils()
+const appStore = useAppStore()
+const vueTorrentStore = useVueTorrentStore()
+
+const sortColumnOptions = computed(() => [
+  { title: t('settings.vuetorrent.searchEngine.sortNone'), value: SearchEngineColumn.NONE },
+  { title: t('searchEngine.headers.fileName'), value: SearchEngineColumn.FILE_NAME },
+  { title: t('searchEngine.headers.fileSize'), value: SearchEngineColumn.FILE_SIZE },
+  { title: t('searchEngine.headers.nbSeeders'), value: SearchEngineColumn.NB_SEEDERS },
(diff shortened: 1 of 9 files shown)


Larsluph (commented):
Pull request template are there for a reason, please don't replace it with an AI generated summary.



Currently, sort is shared across all search tabs. Is it expected to switch to a separated one for each?

Adding a new settings doesn't feel natural, what about persisting the value for reuse across session instead?

Darlio:
Thanks for this — I went looking for exactly this feature and found your PR instead of starting my own, which saved me a lot of time.

I've been running it against qBittorrent 5.2.3 with ~1300-result searches through Jackett. It behaves as described: new tabs inherit the configured sort, and adjusting a tab's sort by hand doesn't write back to the default.

One observation that might be worth a line in the PR description: binding `v-model:sort-by` on the tables also makes the sort persist across re-renders and across new searches within the same tab, not just across new tabs. That's a useful side effect reviewers might not spot from the title.

I've built named search presets on top of your branch — an optional layer for saving named views and marking one as the default. It's cherry-picked onto your six commits with your authorship preserved, and it's purely additive: if presets aren't wanted, nothing in your PR changes.

I'll open it as a separate PR and rebase once yours merges so it doesn't complicate your review. If you'd rather I waited until yours has landed, say so and I'll hold off — yours is the one that needs to go in first, mine only makes sense afterwards.

For transparency: I'm not a Vue developer, and the implementation was AI-assisted (Claude). The testing and measurements are my own.

PR #2867, opened 2026-07-22
feat: show raw data values on hover

### LINKED ISSUE #2427: [Feature Request]: Show raw values when hovering
### Description

For example - torrent sizes are currently showing it in a human readable formats (e.g. GiB).
Could be useful for some cases to be able to hover over the field and see the raw value (bytes in the example above).

_No response_

_No response_

Ceugoeg:
# Raw data values on hover [feat]

## Summary

- show locale-aware raw numeric values in a tooltip while keeping compact values in the UI
- add a reusable `RawNumber` component for dashboard cards and torrent table cells
- add and export `formatRawNumber` with unit tests
- keep tooltip content tied to each rendered value to avoid stale or mismatched hover text

Fixes #2427

## Testing

- `npm run lint`
- `npm test` (25 files, 382 tests)
- `npm run build`

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Core/RawNumber.vue
+++ b/src/components/Core/RawNumber.vue
@@ -18,7 +18,7 @@ const formattedValue = computed(() => {
 <template>
   <v-tooltip :text="formattedValue" location="top">
     <template #activator="{ props: activatorProps }">
-      <span v-bind="activatorProps" class="raw-number">
+      <span v-bind="activatorProps" class="raw-number" tabindex="0">
         <slot />
       </span>
     </template>
@@ -31,5 +31,10 @@ const formattedValue = computed(() => {
   text-decoration-line: underline;
   text-decoration-style: dotted;
   text-underline-offset: 0.2em;
+}
+
+.raw-number:focus-visible {
+  outline: 2px solid currentColor;
+  outline-offset: 2px;
 }
 </style>
--- a/src/components/Core/RawNumber.vue
+++ b/src/components/Core/RawNumber.vue
@@ -1,0 +1,35 @@
+<script setup lang="ts">
+import { computed } from 'vue'
+import { useI18n } from 'vue-i18n'
+import { formatRawNumber } from '@/helpers'
+
+const props = defineProps<{ value: number }>()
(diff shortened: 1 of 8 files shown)


src/components/Core/RawNumber.vue
Larsluph, line 1: Rename to `RawNumberTooltip` for better readability on component usage.

Larsluph (changes requested):
Great!

This component can be used in other places apart from the dashboard.
At least add for Content's TreeNode size.
We'll see later if it needs to be added elsewhere.

Larsluph (approved):

Larsluph:
Thanks for the contribution!

r4ian:
@Ceugoeg is there a way to turn this feature off? I couldn't find it.

r4ian:
@Larsluph maybe you could help us out on how to disable this feature, as we do not need to know the byte-exact size of the torrents, on the other hand this makes the UI more cluttered.

Edit: also:
- there is no unit on the popup value
- the NIST recommendation for grouping digits is not commas but spaces for thousand separators: [https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-10-more-printing-and-using-symbols-and-numbers](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-10-more-printing-and-using-symbols-and-numbers), 10.5.3 paragraph.

image

Larsluph:
@r4ian

I agree now that it's deployed that it create a lot of clutter in the dashboard, I feel like this may only be used in an analysis scenario where you target a single torrent. In this case the full byte size can be displayed in the info tab of the TorrentDetail view only. The tooltip in content view will be left unchanged.

Good point about the unit, I'll add it with all changes.

About the decimal separator, we currently use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat), which uses the US convention for english language, switching to another language will use different rules depending on the locale defaults. We can switch to using static defaults and allowing user to configure them.

PR #2879, opened 2026-08-02
feat: Add ability to define "favorite" download paths

Healdb:
# Feat: add ability to define "favorite" download paths

When downloading things, I often have to write in the specific directory from the default save path that I want to put the files in. This PR allows me to define "favorite" download paths that will always appear in the dropdown menu. Small QoL fix that I find useful.
Defining favorite paths:
favorite1

Paths show in dropdown:

https://github.com/user-attachments/assets/559bdc6c-d9d6-4956-aa37-0a7207e266e4

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/components/Core/ServerPathField.vue
+++ b/src/components/Core/ServerPathField.vue
@@ -9,6 +9,7 @@ import { useAppStore, useHistoryStore } from '@/stores'
 const props = defineProps<{
   title: string
   historyKey: HistoryKey
+  extraItems?: string[]
 }>()
 
 const modelValue = defineModel<string | undefined>({ required: true })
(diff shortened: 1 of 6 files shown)


src/components/Core/ServerPathField.vue
```diff
@@ -21,15 +22,14 @@ const comboboxRef = ref<HTMLInputElement>()
...
-  const combined = [...history]
-  server.forEach(item => {
-    if (!combined.includes(item)) {
-      combined.push(item)
-    }
-  })
+  for (const item of [...(props.extraItems ?? []), ...historyItems.value, ...serverItems.value]) {
+    if (!item || seen.has(item)) continue
+    seen.add(item)
+    combined.push(item)
```
Larsluph, line 31: Not sure why you're using both a set and an array, is it only because sets are not ordered? Why not introduce a sort weight inside the set to sort after all entries have been inserted?

src/components/Settings/Downloads.vue
```diff
@@ -78,6 +79,11 @@ const addStoppedEnabled = computed({
   },
 })
 
+const favoriteSavePaths = computed({
+  get: () => vueTorrentStore.favoriteSavePaths,
+  set: value => vueTorrentStore.setFavoriteSavePaths(value),
+})
+
```
Larsluph, line 86: You can create it directly inside the `vuetorrent` store and only expose this `WritableComputed` instead of the `Ref`

Larsluph (changes requested):

PR #2882, opened 2026-08-02
fix: prevent search field from being cut off on Safari (Vuetify 4 grid layout regression)

### LINKED ISSUE #2805: [Bug]: Search field incomplete
### Bug description

I would like to report a bug, before 2.33.0, everything was OK, now part of the space to search is cut off, I'm going to leave a screenshot. It happens only in Safari on iOS and on macOS,  Firefox is fine. If I revert to prior 2.33.0, everything is fine. 
The latest 2.34.0 also has the Issue.

### Expectations

Image
 space in the search field box

Nil

### Steps to reproduce

1. On Safari only

### Environment description

```
OS: macOS, iOS 
Browser Safari : on iOS and macOS 

PWA: Yes / No
Running inside a container: No
```

### qBittorrent version

5.2.1

### VueTorrent version

2.33.0 and above

### Agreements

waterWang:
## Description

Fixes #2805

The v2.33.0 upgrade from Vuetify 3.x to 4.x introduced CSS grid layout for `v-text-field` components. Safari's WebKit engine doesn't properly inherit `min-width: 0` through flex → grid → flex child chains, causing the search field to shrink below its content width.

## Root Cause

- `v-app-bar` uses flex layout
- `v-text-field` (`.v-input`) uses CSS grid with `grid-template-columns: max-content minmax(0, 1fr) max-content`
- `.v-field` uses CSS grid with `grid-template-columns: min-content minmax(0, 1fr) min-content min-content`
- Safari requires explicit `min-width: 0` on the grid container itself when it's a flex child, unlike other browsers that properly inherit the constraint

## Fix

Added `min-width: 0` to `.v-text-field` in the component's scoped styles, ensuring Safari correctly allows the search field to shrink to fit the available space.

## Testing

- [x] TypeScript check passes (`vue-tsc`)
- [x] Bug reported on Safari iOS and macOS - CSS fix addresses the WebKit rendering difference
- [x] No visual change on Firefox/Chrome (already worked correctly)

### DIFF
--- a/src/components/TorrentSearchbar.vue
+++ b/src/components/TorrentSearchbar.vue
@@ -34,4 +34,12 @@ const torrentTitleFilter = computed({
     @click:clear="resetInput()" />
 </template>
 
-<style scoped></style>
+<style scoped>
+/* Safari fix: Vuetify 4 grid layout needs explicit min-width: 0
+   on the v-text-field to prevent Safari from shrinking the input
+   in flex containers (v-app-bar). This is a known WebKit issue
+   where min-width: 0 isn't properly inherited by grid children. */
+.v-text-field {
+  min-width: 0;
+}
+</style>


src/components/TorrentSearchbar.vue
```diff
@@ -34,4 +34,12 @@ const torrentTitleFilter = computed({
     @click:clear="resetInput()" />
 </template>
 
-<style scoped></style>
+<style scoped>
+/* Safari fix: Vuetify 4 grid layout needs explicit min-width: 0
+   on the v-text-field to prevent Safari from shrinking the input
+   in flex containers (v-app-bar). This is a known WebKit issue
+   where min-width: 0 isn't properly inherited by grid children. */
```
Larsluph, line 41: Can you link a WebKit issue reference here?

PR #2884, opened 2026-08-03
fix(search): guard against overlapping results polls desyncing the offset

### LINKED ISSUE #2883: [Bug]: Search polling has no in-flight guard: slow responses desync the offset into a permanent 409 loop
### Bug description

`SearchEngine.vue` polls `POST /api/v2/search/results` on a fixed 1-second `setInterval`, and `searchEngine.ts` uses `tab.results.length` as the request offset. There is no guard against a poll being issued while the previous one is still in flight.

When a response takes longer than the poll interval — normal for a remote client over WAN/TLS pulling a large result set — two polls overlap. Both read the same `tab.results.length` as their offset, both receive the same window of results, and both `push` it. `tab.results.length` then exceeds the job's `total` on the server, permanently.

From that point qBittorrent answers **every** subsequent poll with `409 Offset is out of range`, and because `refreshResults` has no error handling and the interval callbacks discard their promises (`() => void refreshResults(tab)`), the poller retries once per second **forever**. The page becomes unresponsive while it hammers qBittorrent.

Measured on my own server before diagnosing it:

```
9158 x "POST /api/v2/search/results HTTP/2.0" 409 22
sustained ~480 requests/minute (8 req/s) for 28 minutes
+ 829 x 499 (client aborted) on the same endpoint
```

The 409 body is 22 bytes — qBittorrent's `Offset is out of range`.

Two things make it hard to notice: the failure produces no visible error (the page just goes sluggish), and reloading appears to fix it while re-arming the pollers from persisted session state.

### Expectations

Only one results request in flight per search tab at a time, and polling stops when the job can no longer be served rather than retrying indefinitely.

**1. No in-flight guard — the root cause.** `src/pages/SearchEngine.vue`:

```ts
tab.timer = setInterval(() => void refreshResults(tab), 1000)   // fires regardless of the previous call
```

`src/stores/searchEngine.ts`:

```ts
const response = await qbit.getSearchResults(tab.id, tab.results.length)   // offset read at call time
tab.results.push(...response.results)
```

Two concurrent calls read the same length, fetch the same window, and both push it.

**2. Nothing terminates the loop.** `refreshResults` has no `catch`, and both call sites discard the promise via `void`, so the rejection is swallowed and the interval keeps firing.

**3. A restored tab never stops polling even on success.** `onBeforeMount` calls the store action directly instead of the local wrapper:

```ts
tab.timer = setInterval(() => void searchEngineStore.refreshResults(tab), 1000)
```

so it ignores the returned `'Stopped'` status and never clears its timer. I observed this on nightly: a finished search kept being polled for 11+ minutes at exactly one request per minute — a 1 s interval clamped by background-tab throttling. In a foreground tab that is 1 req/s per restored tab, indefinitely.

**On DevTools:** I did not capture console output for the 409 case, because I was never able to trigger it in a browser (it needs a latency-bound client). The evidence for it is the access log; the reproduction is the unit test. I'd rather say that than present something I didn't observe.

### Steps to reproduce

**Deterministic, no qBittorrent required.** My PR adds `src/stores/searchEngine.spec.ts`. Apply just that file to current `master` and run it:

```console
$ npx vitest run src/stores/searchEngine.spec.ts
FAIL  stores/searchEngine > overlapping polls must not duplicate results
AssertionError: expected 200 to be 100
```

It stands up a fake backend holding 100 results with 20 ms latency and issues two concurrent `refreshResults` calls — exactly what the 1-second interval does when a response takes longer than a second. The client ends up believing it holds 200 results against a server that has 100: the desync that makes every later poll 409.

**In the real world** the conditions are:

1. Access the WebUI from a **remote/high-latency** client (WAN, through a reverse proxy) — not localhost or LAN.
2. Run a search returning a large result set (mine returns ~1300 via the Jackett plugin), so responses are big enough to exceed ~1 s.
3. Leave the Search page open.

I could **not** reproduce it from a LAN client, where the same request completes in milliseconds and polls never overlap. I suspect that's why it hasn't been reported before — it effectively requires a latency-bound client.

Server-side confirmation that an over-large offset is fatal rather than benign, against a live job holding 1347 results:

```console
$ curl ... -X POST /api/v2/search/results --data "id=$ID&offset=99999"
HTTP/1.1 409 Conflict
```

(Note qBittorrent's search jobs are session-scoped — `curl` needs a shared cookie jar, else each request creates a new session and returns 404, not 409.)

Built from source

### Environment description

```plain text
Device type: Desktop
OS: Linux (Ubuntu) server, Docker (linuxserver/qbittorrent); clients on Windows 11
Browser (name and version): Firefox 152 and Chrome 150

PWA: No
Running inside a container: Yes

Behind an nginx reverse proxy - relevant only in that it is the latency path, and that it made the request storm visible in a central access log.
```

### qBittorrent version

5.2.3

### VueTorrent version

2.34.1 (and master @ edd7442)

### Agreements

Darlio:
# Fix overlapping search polls desyncing the offset into a permanent 409 loop [fix]

Fixes #2883

> **Updated 2026-08-04** after @Larsluph's review: polling now goes through the existing `useTimer` composable instead of the hand-rolled guard this PR originally added, and it also fixes the stale-response race CodeRabbit spotted. The sections below describe the current diff.

## What this fixes

See #2883 for the full analysis. In short: `SearchEngine.vue` polls with `setInterval(..., 1000)` and `searchEngine.ts` uses `tab.results.length` as the request offset. There's no in-flight guard, so if a `/search/results` response takes longer than a second, two polls overlap: both read the same offset, both get the same window, and both push it. `results.length` overshoots the job's `total` and never recovers, so qBittorrent returns `409 Offset is out of range` on every subsequent poll — retried forever, because `refreshResults` has no `catch` and both callers use `() => void refreshResults(tab)`.

I measured **9158 x `409` at ~8 req/s over 28 minutes** from a single browser. Notably **8742 of those came from the WAN client and only 416 from the LAN**, plus 829 x `499` (client aborted) — consistent with the latency requirement. I could not reproduce it from a LAN client at all, which is probably why it hasn't surfaced before.

## Changes

**`src/pages/SearchEngine.vue`**
- Replace the per-tab `setInterval` handles with a single page-level `useTimer`. Its task is `.drop()`-ed, so a tick is skipped while the previous one is still in flight — that is the root-cause fix, and it uses the project's existing idiom rather than introducing a new one. It also pauses polling while the VueTorrent tab is hidden.
- One timer for the page rather than one per tab, since a composable can't be instantiated per tab in a loop. The timer refreshes every tab that has a live job and pauses itself when none are left.
- The `onBeforeMount` restore path previously called `searchEngineStore.refreshResults(tab)` directly, ignoring the returned `'Stopped'`, so a restored tab polled forever even after a clean completion. I observed a finished search still being polled 11 minutes later (at 1/min, clamped by background-tab throttling; 1/s in the foreground). It now goes through the same timer.

**`src/stores/searchEngine.ts`**
- Capture the job id before awaiting and discard a response that comes back for a different one. `.drop()` prevents two polls overlapping but not a single in-flight poll outliving the job it was issued for: stop or restart a search while a request is out, and the late response would either append the old job's results to the new job's freshly emptied array — recreating this exact desync — or report a terminal status and tear down the search just started.
- Treat `404` (job gone — qBittorrent keeps search jobs in memory) and `409` (offset past the result count) as terminal, returning `'Stopped'` so the caller stops polling. Other errors propagate unchanged. Narrowed with `axios.isAxiosError()`.

**`src/types/vuetorrent/SearchData.ts`**
- Remove the now-unused `timer` field. Flagging this as scope creep on my part — happy to leave it in place if you'd rather.

**`src/stores/searchEngine.spec.ts`** (new)
- Four tests: a stale success response is discarded, a stale error doesn't stop the new job, and `404` / `409` are terminal. All four fail against the unpatched store and pass against the patched one.
- This is the first spec under `src/stores/`, so I kept it to plain `vitest` + `createPinia()` with no new helpers or patterns.

## Testing

- New spec against the unpatched store: **4/4 fail**. With the fix: **4/4 pass**
- Full suite: **380 tests / 26 files** (376 on master + 4 here), all passing
- `npm run lint` — clean · `vue-tsc` — clean · `npm run build` — clean
- Running the patched build against qBittorrent **5.2.3** on my own server. A full search polled at exactly 1 req/s across 106 consecutive seconds, never 2, then stopped cleanly, with 0 x 409. *(That measurement is from the original implementation; the current one polls through `useTimer`, which is strictly more conservative.)*

## Behaviour changes worth flagging

- A search that hits `409`/`404` now **stops** rather than retrying forever. That's the intent, but the search ends quietly with no message to the user. A toast would arguably be better; I left it out to keep the diff small and to avoid inventing copy and a translation key uninvited. Happy to add it.
- Polling now pauses while the tab is hidden, inherited from `useTimer`. A background search collects results only once the tab is focused again.

## Limitations — what I did *not* verify

- The tests cover the store. Nothing covers the timer wiring itself, which would need the page mounted with router, i18n, Vuetify and several stores — happy to write that if you want it.
- I never reproduced the `409` storm in a browser — only in the unit test and in production logs. The real-world path needs a latency-bound client.
- Tested against qBittorrent 5.2.3. Per @Larsluph's suggestion I traced `searchcontroller.cpp` instead of testing older builds: the `404` / `409` paths this PR relies on are unchanged from `release-4.2.5` to current master (details in the comments below).

## PR Checklist

- [x] All tests pass

---

*The investigation and patch were done with AI assistance (Claude); I reviewed the diff, built and deployed it, and the measurements above are from my own instance.*

### DIFF
--- a/src/pages/SearchEngine.vue
+++ b/src/pages/SearchEngine.vue
@@ -5,7 +5,7 @@ import { useRouter } from 'vue-router'
 import { useDisplay } from 'vuetify'
 import HistoryField from '@/components/Core/HistoryField.vue'
 import PluginManagerDialog from '@/components/Dialogs/PluginManagerDialog.vue'
-import { useI18nUtils, useSearchQuery } from '@/composables'
+import { useI18nUtils, useSearchQuery, useTimer } from '@/composables'
 import { HistoryKey } from '@/constants/vuetorrent'
 import { comparators, formatData, formatTimeSec, openLink } from '@/helpers'
 import { useAddTorrentStore, useAppStore, useDialogStore, useSearchEngineStore, useVueTorrentStore } from '@/stores'
@@ -93,17 +93,30 @@ function openResultLink(result: SearchResult) {
   openLink(result.descrLink)
 }
 
+// One timer for the page rather than one interval per tab: useTimer drops a run while the
+// previous one is still going, so a response slower than the interval can no longer be raced
+// by the next tick — which is what let two polls read the same offset and push the same
+// results twice, overshooting the server's count and 409-ing every poll after that.
+const { pause: pausePolling, resume: resumePolling } = useTimer(refreshActiveSearches, 1000, { immediate: false })
+
+async function refreshActiveSearches() {
+  const activeTabs = searchData.value.filter(tab => tab.id !== 0)
+  if (activeTabs.length === 0) {
+    pausePolling()
+    return
+  }
+
+  await Promise.allSettled(activeTabs.map(refreshResults))
+}
+
 async function runNewSearch() {
   await searchEngineStore.runNewSearch(selectedTab.value)
-  selectedTab.value.timer = setInterval(() => void refreshResults(selectedTab.value), 1000)
+  resumePolling()
   queryInput.value?.saveValueToHistory()
 }
 
 async function stopSearch(tab: SearchData) {
   await searchEngineStore.stopSearch(tab)
-  if (tab.timer) {
-    clearInterval(tab.timer)
-  }
 }
 
 function stopAllSearch() {
@@ -138,21 +151,16 @@ onBeforeMount(async () => {
   document.addEventListener('keydown', handleKeyboardShortcut)
   if (searchData.value.length === 0) {
     searchEngineStore.createNewTab()
-  } else {
-    searchData.value.forEach((tab: SearchData) => {
-      if (tab.id && tab.id !== 0) {
-        tab.timer = setInterval(() => void searchEngineStore.refreshResults(tab), 1000)
-      }
-    })
+  } else if (searchData.value.some((tab: SearchData) => tab.id !== 0)) {
+    // A tab restored from sessionStorage may still reference a live job. Resume so it keeps
+    // collecting results, and so it stops on its own once the job reports 'Stopped'.
+    resumePolling()
   }
   await searchEngineStore.fetchSearchPlugins()
 })
 
 onBeforeUnmount(() => {
   document.removeEventListener('keydown', handleKeyboardShortcut)
-  searchData.value.forEach((tab: SearchData) => {
-    if (tab.timer) clearInterval(tab.timer)
-  })
 })
 </script>
 
--- a/src/stores/searchEngine.ts
+++ b/src/stores/searchEngine.ts
@@ -1,3 +1,4 @@
+import { isAxiosError } from 'axios'
 import { acceptHMRUpdate, defineStore } from 'pinia'
 import { v4 as uuidv4 } from 'uuid'
 import { ref } from 'vue'
@@ -24,7 +25,6 @@ export const useSearchEngineStore = defineStore(
           plugin: 'enabled',
         },
         results: [],
-        timer: null,
       })
     }
 
@@ -40,10 +40,33 @@ export const useSearchEngineStore = defineStore(
     }
 
     async function refreshResults(tab: SearchData) {
-      const response = await qbit.getSearchResults(tab.id, tab.results.length)
-      tab.results.push(...response.results)
+      // The job can be restarted or stopped while this request is in flight. Remember which
+      // job we asked about so a late response can't be applied to a different one.
+      const requestedId = tab.id
 
-      return response.status
+      try {
+        const response = await qbit.getSearchResults(requestedId, tab.results.length)
+        if (tab.id !== requestedId) {
+          // Stale: appending here would corrupt the new job's offset, and reporting a
+          // terminal status would stop a search the user has just started.
+          return 'Running' as const
+        }
+        tab.results.push(...response.results)
+
+        return response.status
+      } catch (error) {
+        if (tab.id !== requestedId) {
+          return 'Running' as const
+        }
+        // 404: the search job no longer exists (qBittorrent keeps them in memory, so a restart drops them).
+        // 409: our offset is beyond the job's result count, so every subsequent poll fails the same way.
+        // Neither can recover by retrying, so report the job as stopped and let the caller stop polling.
+        if (isAxiosError(error) && (error.response?.status === 404 || error.response?.status === 409)) {
+          return 'Stopped' as const
+        }
+
+        throw error
+      }
     }
 
     async function stopSearch(tab: SearchData) {
--- a/src/stores/searchEngine.spec.ts
+++ b/src/stores/searchEngine.spec.ts
@@ -1,0 +1,116 @@
+import { createPinia, setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, test, vi } from 'vitest'
+import { useSearchEngineStore } from './searchEngine'
+
+/**
+ * Simulated qBittorrent search backend.
+ *
+ * Mirrors searchcontroller.cpp: results accumulate while a job runs, and an
+ * offset past the current result count is rejected with 409 Conflict
+ * ("Offset is out of range").
+ */
+const server = {
+  total: 0,
+  latencyMs: 0,
+  /** Simulates a job qBittorrent no longer knows about (it keeps them in memory). */
+  jobMissing: false,
+}
+
+/** axios.isAxiosError() only checks this flag, so the store treats these as real API errors. */
+function apiError(message: string, status: number) {
+  return Object.assign(new Error(message), { isAxiosError: true, response: { status } })
+}
+
+vi.mock('@/services/qbit', () => ({
+  default: {
+    getSearchResults: vi.fn(async (_id: number, offset = 0) => {
+      // Offset is captured by the caller *before* this resolves — that window is the bug.
+      await new Promise(resolve => setTimeout(resolve, server.latencyMs))
+
+      if (server.jobMissing) {
+        throw apiError('Not Found', 404)
+      }
+
+      if (offset > server.total) {
+        throw apiError('Offset is out of range', 409)
+      }
+
+      const results = Array.from({ length: server.total - offset }, (_, i) => ({ fileName: `result-${offset + i}` }))
+      return { results, status: 'Running' as const, total: server.total }
+    }),
+    stopSearch: vi.fn(() => Promise.resolve()),
+    startSearch: vi.fn(() => Promise.resolve({ id: 1 })),
+    getSearchPlugins: vi.fn(() => Promise.resolve([])),
+  },
+}))
+
+function makeTab(store: ReturnType<typeof useSearchEngineStore>) {
+  store.$reset()
+  const tab = store.searchData[0]
+  tab.id = 1
+  tab.results = []
+  return tab
+}
+
+describe('stores/searchEngine', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    server.total = 0
+    server.latencyMs = 0
+    server.jobMissing = false
+  })
+
+  test('a response for a job that is no longer current is discarded', async () => {
+    const store = useSearchEngineStore()
+    const tab = makeTab(store)
+
+    server.total = 100
+    server.latencyMs = 20
+
+    const inFlight = store.refreshResults(tab)
+    // The user restarts the search while that request is still out: runNewSearch() swaps in
+    // the new job id and empties the results.
+    tab.id = 2
+    tab.results = []
+
+    // Appending the old job's 100 results here would put the new job's offset past its own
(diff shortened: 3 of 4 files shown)


Larsluph:
> Tested against qBittorrent 5.2.3 only; I haven't checked older WebUI API versions for different status codes on these paths.

You don't really have to test it but look at code changes inside the search engine controller, there's a high chance things have changed since the time I developed this feature.

[qbittorrent Search Engine controller source](https://github.com/qbittorrent/qBittorrent/blob/master/src/webui/api/searchcontroller.cpp)

src/pages/SearchEngine.vue
```diff
@@ -93,16 +93,26 @@ function openResultLink(result: SearchResult) {
...
 
+function startPolling(tab: SearchData) {
+  // Always clear first: assigning over a live handle orphans the previous interval,
+  // which then keeps polling for the lifetime of the page.
+  if (tab.timer) {
+    clearInterval(tab.timer)
+  }
+  tab.timer = setInterval(() => void refreshResults(tab), 1000)
+}
+
```
Larsluph: Intervals have been replaced with `useTask` for quite some time now, it is  ensuring not to have concurrent requests, I probably forgot about this one when refactoring everything.

I recently built a composable around it to prevent network usage when the VueTorrent tab is not focused, look what you have to do to use it.

`useTimer`: src/composables/Timer.ts

Larsluph (changes requested):

Darlio:
Thanks — that was better advice than testing versions, and a lot quicker.

I traced `searchcontroller.cpp` back through its history. The two error paths this PR relies on are unchanged from `release-4.2.5` through to current master:

```cpp
if (iter == m_searchHandlers.end())
    throw APIError(APIErrorType::NotFound);                                  // 404

if (offset > size)
    throw APIError(APIErrorType::Conflict, tr("Offset is out of range"));    // 409
```

(the same `Conflict` also covers a negative offset after normalisation)

The commits touching that file since, and why none of them change the picture:

- `0c48b70` (2025-06-05) *Send 204 when WebAPI response contains no data* — the one I was most concerned about, but it doesn't reach here: every success path of `resultsAction()` returns a populated object (`status` / `results` / `total`). Had it applied, this PR would have been wrong — a 204 would leave `response.results` undefined and `push(...response.results)` would throw straight back into the same retry loop.
- `10b879b` (2025-09-27) *Propagate error messages from search engine* — logging only
- `dae1fdf` (2026-06-14) type renames and `eed0e56` (2025-09-28) return types — no behavioural change

So the `404`/`409` pair holds across every qBittorrent version VueTorrent supports, not just the 5.2.3 I happened to test on. Happy to update the PR description to say that instead of the narrower caveat — or to put a short comment next to the check pointing at the controller, if you'd rather the reasoning lived in the code. Your call, I'll do whichever you prefer.

One unrelated thing I noticed while I was in there, purely FYI and not part of this PR: `statusAction()` with `id=0` and no active searches returns an empty array, which is exactly the shape that 204 change targets. If it ever does start returning 204 there, `getSearchStatus()` would receive an empty body. Might be worth a glance at some point.

Darlio:
Thanks — both points were right, and the result is smaller than what I had.

**`useTimer` instead of intervals.** I'd missed the composable entirely. Beyond dropping concurrent runs, the visibility pause matters for this bug specifically: the runaway I hit ran at ~8 req/s for 28 minutes before anyone noticed. I can't prove the tab was hidden for all of that, but a poller that stops when nobody's looking bounds this class of runaway regardless.

Using it meant going from one interval per tab to **one timer for the page**, since a composable can't be instantiated per tab in a loop. The timer walks the tabs that have a live job, refreshes each, and pauses itself when none are left. Consequences:

- `startPolling()` is gone, and with it the orphaned-interval case it was working around
- `SearchData.timer` is now unused, so I removed the field — **flagging that as scope creep on my part**; say the word and I'll put it back
- my hand-rolled in-flight `Set` is gone, superseded by `.drop()`

**The stale-response race CodeRabbit found is real, and my own review missed it.** `.drop()` stops two polls overlapping, but not a single in-flight poll outliving the job it was issued for: stop or restart a search while a request is out, and the late response either appends the old job's results to the new job's empty array — recreating the exact desync this PR is about — or returns `Stopped` and tears down the search that was just started. So the store now captures the job id before awaiting and discards anything that comes back for a different one. Two tests cover it.

Also swapped the `(error as AxiosError)` cast for `axios.isAxiosError()`, which is what made me notice the spec's fake errors weren't axios-shaped at all.

**Validation** (Node 20.20.1): `eslint`, `vue-tsc` and `npm run build` clean; **380 tests / 26 files** (376 on master + 4 here). I checked the four new tests fail against the unpatched store before they pass against the patched one, so they aren't tautological. It's running on my own instance against qBittorrent 5.2.3.

**One gap I'd rather name than paper over:** the tests cover the store; nothing covers the timer wiring itself, since that needs the page mounted with router, i18n, Vuetify and several stores. Happy to write it if you want it — I just didn't want to add a heavy component spec uninvited.

src/stores/searchEngine.ts
```diff
@@ -40,10 +44,34 @@ export const useSearchEngineStore = defineStore(
...
+        if (status === 404 || status === 409) {
+          return 'Stopped' as const
+        }
 
-      return response.status
+        throw error
+      } finally {
+        pendingRefresh.delete(tab.uniqueId)
+      }
     }
```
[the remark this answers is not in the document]
Darlio: Good catch — this was real, and my own review had missed it.

The `useTimer` migration @Larsluph asked for gives `.drop()`, which stops two polls overlapping — but not a single in-flight poll outliving the job it was issued for, which is exactly the case you describe.

Fixed in `dc35652`: `refreshResults` captures `requestedId = tab.id` before the await and discards the response if `tab.id` has changed, in **both** the success and error paths — so a late response can neither push the old job's results into the new job's freshly emptied array, nor report a terminal status that tears down the search just started.

Two regression tests added (*a response for a job that is no longer current is discarded*, *an error for a job that is no longer current does not stop the new job*); both fail against the unpatched store.

Your comment also made me notice the spec's fake errors weren't axios-shaped at all, so the error path is now narrowed with `axios.isAxiosError()` rather than a cast.

Resolving — happy to reopen if you'd rather it were handled differently.

src/pages/SearchEngine.vue
Darlio: Thanks for pointing at it — I hadn't found `useTimer` and would have shipped a worse version of something you already have.

Done in `dc35652`. One consequence worth flagging: a composable can't be instantiated per tab in a loop, so this went from one interval per tab to **one page-level timer** that refreshes each tab with a live job and pauses itself when none are left. That removed `startPolling()` and made `SearchData.timer` dead, so I dropped the field — say the word if you'd rather keep it.

The visibility pause is a real win here beyond the concurrency guard: the runaway I originally hit ran unattended at ~8 req/s, and that alone would have bounded it.

Larsluph:
> One unrelated thing I noticed while I was in there, purely FYI and not part of this PR: `statusAction()` with `id=0` and no active searches returns an empty array, which is exactly the shape that 204 change targets. If it ever does start returning 204 there, `getSearchStatus()` would receive an empty body. Might be worth a glance at some point.

This might be worth noting. Using only 200 checks is good but not future-proof, especially now that we know can change in the future.


> I can't prove the tab was hidden for all of that, but a poller that stops when nobody's looking bounds this class of runaway regardless.

Just to make sure it is clear, the visibility change API fire events only on tab change, not when the browser window loses focus with the VueTorrent tab focused. This may depend on the browser but it's what I got on chrome.

Will review code later...

src/pages/SearchEngine.vue
```diff
@@ -93,17 +93,30 @@ function openResultLink(result: SearchResult) {
...
 }
 
+// One timer for the page rather than one interval per tab: useTimer drops a run while the
+// previous one is still going, so a response slower than the interval can no longer be raced
+// by the next tick — which is what let two polls read the same offset and push the same
+// results twice, overshooting the server's count and 409-ing every poll after that.
+const { pause: pausePolling, resume: resumePolling } = useTimer(refreshActiveSearches, 1000, { immediate: false })
+
+async function refreshActiveSearches() {
+  const activeTabs = searchData.value.filter(tab => tab.id !== 0)
```
Larsluph, line 103: This lambda is used twice, it can be extracted.

Larsluph (changes requested):
conflicts, please update

WDaan (approved):

PR #2912, opened 2026-08-31
fix: category form not fetching incomplete torrent path details

### LINKED ISSUE #2905: [Bug]: Per-category incomplete torrent path not restored correctly
### Bug description

You can configure a custom path for incomplete torrents when editing a category. This configuration is saved, but does not get reloaded when revisiting the Edit Category web UI.

### Expectations

The checkbox state and configured path should be loaded correctly.

In categories.json, there appears to be a single "download_path" property to represent this data.

### Steps to reproduce

1. Visit Settings / Tags & Categories
2. Edit a category and check the "Use another path for incomplete torrents" checkbox.
3. Enter a custom path
4. Save changes
5. Edit the category again and observe that the checkbox is now unchecked and the configured path is missing.

### Environment description

```plain text
Device type: Desktop
OS: Windows
Browser: Chrome 151.0.7922.170

PWA: No
Running inside a container: linuxserver Docker
```

### qBittorrent version

v5.2.3

### VueTorrent version

v2.34.1

### Agreements

jagadam97:
# category form not fetching incomplete torrent path details while form renders in edit mode [fix]

for Categories form in settings the incomplete download path is not populating correct as we are currently not transforming the data of downloadPathEnabled and downloadPath from download_path. where as for syncFromMetadata and other functions we are actually transforming the data to 2 keys.

This PR fixes this but correctly transforming data for Categories while loading them into store.

closes #2905

Proof of Testing:
https://github.com/user-attachments/assets/4f87118a-7658-40eb-b7cd-341070efd54d

## PR Checklist

- [x] All tests pass

### DIFF
--- a/src/stores/categories.ts
+++ b/src/stores/categories.ts
@@ -10,7 +10,12 @@ export const useCategoryStore = defineStore('categories', () => {
   /** Key: Category name */
   const _categoryMap = shallowRef<Map<string, Category>>(new Map())
   const categories = useSorted(
-    () => Array.from(_categoryMap.value.values()),
+    () =>
+      Array.from(_categoryMap.value.values()).map(cat => ({
+        ...cat,
+        downloadPathEnabled: cat.download_path !== undefined && cat.download_path !== false,
+        downloadPath: cat.download_path !== undefined && typeof cat.download_path === 'string' ? cat.download_path : '',
+      })),
     (a, b) => comparators.text.asc(a.name, b.name)
   )
 
--- a/src/stores/categories.ts
+++ b/src/stores/categories.ts
@@ -10,12 +10,7 @@ export const useCategoryStore = defineStore('categories', () => {
   /** Key: Category name */
   const _categoryMap = shallowRef<Map<string, Category>>(new Map())
   const categories = useSorted(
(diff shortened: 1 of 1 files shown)


Larsluph:
What's the root cause of this, is it because of a breaking change in qbit's api response?

`download_path` being mapped to `downloadPathEnabled` and `downloadPath`?
The mapping should already be made in the maindata sync, why does data not get passed around?

jagadam97:
with my changes the initial props the form is getting was without downloadPath or DownloadPathEnabled
Screenshot 2026-09-02 at 10 04 01 PM
Screenshot 2026-09-02 at 10 04 43 PM


because when full update is true at https://github.com/VueTorrent/VueTorrent/blob/master/src/stores/categories.ts#L37
we are not mapping `downloadPathEnabled` and `downloadPath` from `download_path` as it just maps entries without any transformation
Screenshot 2026-09-02 at 10 09 39 PM


i taught fix would be better if we are loading categories with proper data but in hind sight better fix would be to fix syncFromMetadata.

 will make the change and push the same in new commit

WDaan (approved):

